PR #29, opened 2015-10-30
Added dependency resolution to make files.

rbooth200:
Add header dependency resolution by generating extra rules using gcc's header resolution options.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -282,3 +282,33 @@ clean ::
 	\rm -f ../analysis/swig_generated/*.so
 	\rm -f *.pyc
 	\rm -f ../analysis/swig_generated/SphSim.py
+	\rm -f $(PWD)/src/*.d
+
+
+# Handle the dependencies using black magic
+get_dir_str 	= $(patsubst %/,%,$(dir $1))
+WHICH_GPP = which g++
+ifeq ($(findstring g++,$(WHICH_GPP)),g++)
+BUILD_DEPS = @set -e; g++ -M -MG $< |\
+	 sed 's;\($*\)\.o[ :]*;$(call get_dir_str, $@)/\1.o $@ : ;g' > $@ 
+else
+BUILD_DEPS = 
+endif
+
+CPP_TGT := $(patsubst %.o,  %.cpp, $(OBJ)) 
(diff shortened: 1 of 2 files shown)


src/Makefile
```diff
@@ -282,3 +282,33 @@ clean ::
...
+WHICH_GPP = which g++
+ifeq ($(findstring g++,$(WHICH_GPP)),g++)
+BUILD_DEPS = @set -e; g++ -M -MG $< |\
+	 sed 's;\($*\)\.o[ :]*;$(call get_dir_str, $@)/\1.o $@ : ;g' > $@ 
+else
+BUILD_DEPS = 
+endif
+
+CPP_TGT := $(patsubst %.o,  %.cpp, $(OBJ)) 
+CPP_TGT := $(patsubst chealpix.cpp, , $(CPP_TGT))
```
giovanni-rosotti: What does healpix have to do with this problem? Is that name just a placeholder?

rbooth200: patsubstr is using the list of objects to generate a .cpp file name. The dependency resolution checks to see if these files 1) exist and 2) have been modified. Since chealpix is a .c file, Make obviously won't find chealpix.cpp as it doesn't exist. Therefore, I just use patsubstr to remove it from the list. 

It would be neater to create a list of .cpp files and .c files that I could use for dependency resolution, from which the .o file list is created. But this method works and required a much smaller change.

giovanni-rosotti: ok, fine. Well, a simple search tells me that chealpix is the only .c file. So it's probably fine for now (unless we're going to add more .c files). I'll test after lunch if the dependencies work also on the mac

giovanni-rosotti:
Ok, works for me on the mac too. Two last points before merging (I'm being very picky just because the makefile is such an important part of the interface with the user):
The dependency files contain also system header files - is that we want? I suppose it shouldn't be a problem, just make will take a bit longer because it needs to check also those timestamps (hardly a problem).
I wonder if we should use the CPP variable rather than g++ (but checking if g++ is part of the name) in case the user does not have a command called g++. Or maybe we should just add a warning in this case informing the user that we were not able to compute the dependencies.

rbooth200:
In which case, replace -M with -MM

rbooth200:
It only reports files included using "" and not <>

PR #36, opened 2015-11-05
Fixes to meshless scheme (for block timesteps), usergiude and other small changes

dhubber:
1. Fixed meshless scheme for block timesteps.  Should work for open and periodic boundaries (unsure about mirror boundaries but since this is possibly being changed soon, I'll leave this for now)
2. Userguide has been updated to include section on units
3. Some other smaller changes, including fixing a small potential problem with scaling the sink density

### DIFF
--- a/docs/userguide.tex
+++ b/docs/userguide.tex
@@ -1626,15 +1626,15 @@ The \var{Simulation} class is the class that contains every other class, includi
 \item \var{AlgorithmSimulation} - Child class containing specific algorithmic implementation.  For example, `grad-h' SPH simulations are run by the \var{GradhSphSimulation} class; pure N-body simulations are run by the \var{NbodySimulation} class, etc..
 \end{itemize}
 
-\subsubsection{Sph class}
+\subsubsection{Hydrodynamics class}
 
-\subsubsection{SphKernel class}
+\subsubsection{SmoothingKernel class}
 
-\subsubsection{SphIntegration class}
-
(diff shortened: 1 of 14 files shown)


giovanni-rosotti:
I think this is fine and you can go ahead - at least it compiles. For the future, please split it in smaller chuncks though. And with more informative comments than "some changes" :)
I see you corrected some problem in UpdateWorkCounters - I was definitely having crashes inside there, so that's good... can't comment on the changes with the meshless. 
I see that there is now code to initialise the sinks with the meshless - good. One small detail you might want to check is that the initialisiation is done in the right order with MPI (I think the sinks class need mpicontrol, which therefore must be allocated before the sinks - and then sinks must be given a pointer to it). To be fair the meshless probably doesn't have any MPI code in it - but we can start putting down the infrastructure to make our life easier when MPI comes.
Last point: I think there are better ways to align latex equations than using manual spacing, but will do for now

dhubber:
Yes, that's what I said before.  I didn't know how the pull requests worked and hoped you could push individual commits but actually it only allowed me to merge the whole branch in one go so will remember that for the future.  There's still some sink/N-body code to be added to the meshless so it's not complete yet but hopefully in the next week or so.  And yes, I've commented out/not included any MPI in the meshless for now, but in principle there are no serious differences that will hinder the meshless with MPI.  About the latex, I just copied and pasted from my talk latex files.  If it looks okay, I am not worrying about it :-)

PR #38, opened 2015-11-06
Hybrid

giovanni-rosotti:
A few changes to make the code compile in hybrid mode (MPI+OpenMP). One change affects also serial code (intel and gcc compiler behave differently when using a const variable inside a parallel region; the easiest way I found to avoid becoming crazy was just to remove the const...)

### DIFF
--- a/src/Common/IC.cpp
+++ b/src/Common/IC.cpp
@@ -3799,7 +3799,7 @@ void Ic<ndim>::AddSinusoidalDensityPerturbation
 
   // Loop over all required particles
   //-----------------------------------------------------------------------------------------------
-#pragma omp parallel for default(none) shared(kwave,r) private(diff,i,xnew,xold)
+#pragma omp parallel for default(none) shared(lambda,Npart,amp,kwave,r) private(diff,i,xnew,xold)
   for (i=0; i<Npart; i++) {
     xnew = r[ndim*i];
 
--- a/src/Common/IC.cpp
+++ b/src/Common/IC.cpp
@@ -3782,9 +3782,9 @@ void Ic<ndim>::AddAzimuthalDensityPerturbation
 //=================================================================================================
 template <int ndim>
 void Ic<ndim>::AddSinusoidalDensityPerturbation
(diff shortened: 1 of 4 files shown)


rbooth200:
Is the hybrid currently working (to some degree) ?

giovanni-rosotti:
As far as I know, yes - I run a few tests and they were fine. The only problem I found is when using a lattice in the initial conditions - but as we have discussed this is more likely to be a problem in that routine rather than in the hybrid parallelization...

rbooth200:
So there is a problem with the build at the moment when mpic++ invokes icc, since the -ffast-math flag and -openmp flags are different.

giovanni-rosotti:
Well there's nothing we can do in this case - we can't understand from the command if you're using gnu or intel given that it's always mpic++. 
There are only two solutions:
* the user modifies manually the flags. Not the end of the world given that you need to do it only once
* we add another variable in the Makefile specifying which compiler you have

rbooth200:
I've pushed some changes that should fix the names (I'd forgotten that I was playing around with them). I've also fixed the timing names, which arose because I got rid of a load of functions in GradhSphTree.cpp

rbooth200:
Not without some shell magic anyway, which is difficult to make portable. Do you have a preference for which solution is used? Perhaps the easiest thing is to remove the automatic choice of compiler flags and to give a few sensible examples instead.

rbooth200:
So I've added the change to the src/Makefile. I'd happily push this except that the travis checks have failed. Although it compiles fine for me. Do you expect it to work with travis?

giovanni-rosotti:
Well, it fails because there is no travis script in this branch... let's first merge this with master (but it should be David that accepts this request as by our policy). I'll then create a request for merging the branch where I set up travis and then we should be fine. Until that is done, expect all the other pushes to fail...

dhubber:
Okay, I just pulled and double checked this branch was working, which it was on the tests I run.  I had the same issue when I tried to send a pull request (for the load balancing) so hopefully we can get this set-up asap.

PR #39, opened 2015-11-09
Added section in userguide describing data file formats

dhubber:
Tried to describe the column and seren file formats while being clear and not overly verbose.  For some of the more complex elements (e.g. the sink particle data), I've suggested the reader looks at parts of the source code (not unreasonable for an advanced user in an open source project I feel).  I might need to add a little more information here and there, but for now I think this will be sufficient to help those who want to understand the basics of both formats.

### DIFF
--- a/docs/userguide.tex
+++ b/docs/userguide.tex
@@ -1007,7 +1007,7 @@ kdtree & = Use simple KD-tree decomposition
 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\section{Input and output file formats}
+\section{Input and output file formats} \label{S:IO}
 
 GANDALF supports several simple file formats for reading in initial conditions or outputting snapshots for visualisation and analysis.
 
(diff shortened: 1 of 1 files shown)


giovanni-rosotti:
I think it's basically fine. However, there are two things missing I believe:
a) some details about the size of the header. Looking at the source code is fine for the sinks as you don't always need them, but the header is something that you need to parse if you want to open the file. And you can't even skip it and jump directly to the data because the length partially depends on its content, so you need to parse at least part of it before you can jump to the data section. While one can still guess the length of the numbers (int 4 bytes, double 8, ...), there's no way to guess the length of the strings...
b) clarify that vector quantities are written together (i.e. x y z for a given particle, then x y z for the next one, ...), and they are not treated as ndim different scalar quantities

PS I don't find a routine for reading the lite format, am I missing something?

dhubber:
Yes, I was aware that perhaps it was lacking a little in descriptive details so I will try and add this extra info on the next iteration.  Regarding the lack of a routine for reading the lite format, that is by design.  It is only supposed to be a small output format designed for producing movies and is useless for restarts because it does not have enough information (such as velocities) for a full simulation restart.  Therefore it does not have a read function to prevent users from trying to use them to restart simulations (which would give rubbish results)

giovanni-rosotti:
I understand that, but how can we produce movies if we can't even read them? Or is SPLASH able to read them?

dhubber:
Ah ok, now I see where you're coming from!  I believe the 'lite' format was added before the 'make_movie' function existed so the only way we had to create movies was via splash.  But you're right that we should be able to read them for making movies.  In principle, we might have been able to use the 'ReadSerenUnformSnapshotFile' routine directly except the lite snapshot defaults to single precision so if the code were compiled in double precision then I guess it wouldn't work.  However, maybe there's a way to use it to read in single precision also with the same code or is that too complicated a change?  However, if that is a possibility, we should add some kind of safeguard that the file is only read in so that it can be copied to the snapshot class and not used for running simulations.

rbooth200:
I think that modifying  ReadSerenUnformSnapshotFile should be relatively straightforward, as long as you can tell from the header the precision used/ It would require templating the reading function on the floating type and calling the appropriate one based on header info.

dhubber:
I've just added a little more info on the headers (only the bare minimum) plus some details on settting up ICs in C++ (again only the minimum details the user needs)

giovanni-rosotti:
I just pushed a small change to clarify how vector quantities are written. About the seren lite format, I'd suggest creating an issue for that. I think we can then accept this pull request as long as you are fine with my changes...

PR #43, opened 2015-11-10
Installing the code

giovanni-rosotti:
I added the setup.py file which allows one to install gandalf in a system folder (and yes, if you do that your folder doesn't need to be called "gandalf" anymore!).
I also added two new targets to the makefile - install and installpython. The latter just executes setup.py. The first one instead just copies the gandalf executable to /usr/local/bin (which normally is already in the PATH). There's probably a better way of doing it, but it should work...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,3 +78,8 @@ clean:
 depends:
 	@+$(MAKE) depends -C src
 
+installpython: all
+	./setup.py install
+	
+install: executable
+	cp bin/gandalf /usr/local/bin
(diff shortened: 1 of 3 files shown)


rbooth200:
This sounds fine to me. I slightly object to installing a simulation code into system directories, but it doesn't make it the wrong thing to do

giovanni-rosotti:
I've updated also the documentation relative to this problem. I think it should now be ready to be merged

dhubber:
Okay, I just tried it out and works fine for me also (both python and executable builds).  As long as it's not the default make option, and it requires root access (which it does), then I'm fine with it.

rbooth200:
When I tried it, it also worked fine. I think it might be worth adding a discussion about using .pydistutils.cfg to specify the install location etc.

PR #44, opened 2015-11-11
Travis

giovanni-rosotti:
This is the branch where I've setup travis. The hybrid openMP+MPI build fails, but that's because the branch travis.

Note that I needed a few modifications to the Makefile to make it work with Travis and be able to use the various combinations of parameters (I can't edit directly the Makefile on Travis...). 
While doing that I discovered that make automatically defines the variable CXX inside a Makefile. So I removed a default value for CPP - if that field is blank I use the default c++ compiler provided by the makefile. If one wants can modify the Makefile and give an explicit value. It doesn't have to be this way, so I'd like to know what you prefer


rbooth200:
Is it worth merging this with the hybrid branch first and then master when they both work together?

giovanni-rosotti:
Well, actually you can already see that this works fine with the changes in hybrid. The last push build failed, but the pull request build succeeded, because after this branch will be merged in master there will be both the modifications here and in hybrid...

rbooth200:
Ok, great. I hadn't completely understood the differences between the two travis builds, but it makes sense...

PR #45, opened 2015-11-12
Loadbalancing

dhubber:
Fixed some issues with load balancing, in particular iterating from -/+ infinity when determining the mpi tree divisions.  Also fixed some small problems that prevented Sph simulations running with Nhydro = 0.  Finally, I merged with the travis branch hoping it would be green-lit, but still seems to be some problem.


giovanni-rosotti:
Don't worry about travis. If you see the pull request build (that is, how the code will be after this will have been merged into master) succeeds. What fails is your last push - but the reason why it fails is because the hybrid branch has not been merged inside this one (if you open the log you see that only the hybrid openMP-MPI build fails, the others succeed), while it has already been merged inside master. So it's not your fault.

It looks reasonable to me - I'll play a bit with it...

giovanni-rosotti:
I don't see any problems with it, I'm merging. On the long run we might want to use something that converges faster than bisection - but for now it will do.

Issue #42, opened 2015-11-10
Analytical solutions


rbooth200:
The userguide mentions that some additional analytical solutions would be nice. I already have a python script that solves the sedov problem, which might be useful. I've attached it here in case.
[SedovSolution.py.txt](https://github.com/gandalfcode/gandalf/files/31041/SedovSolution.py.txt)

dhubber:
The Sedov solution has already been integrated into the Python script and works perfectly fine.  Any additional analytical solutions can be added as needed into the same file in the future.

PR #49, opened 2015-11-16
Added sedov solution to the analytical solutions

rbooth200:
I've added the sedov solution to the list of analytical solutions, and corrected a couple of bugs in the file that I included in issue #42. This solution does add a scipy dependency, but it's only really being used for interpolation. Is this ok? 

Also, in doing this I realised that it would be good to have a dimension independent way to get things like |x| and |v| (in python). Is this possible/reasonable?


giovanni-rosotti:
scipy is not a problem - we're already using interpolate in other places. I'm not concerned - if you have numpy and matplotlib, in 99% cases you have also scipy.

you mean the magnitude of the vectors? yeah, i guess i could add a function that computes that automatically by retrieving ndim. and then pre-define at the beginning the lengths of x and v

rbooth200:
Well something like that. It's just that you want to plot r vs rho & v etc in the analytical solution to the sod problem, which can be done in any number of dimensions. It made me realise that there is currently no sensible way to do this.

dhubber:
Okay, I just tried it and it mainly works fine.  Seems to plot the correct values for rho, u and P.  I'm guessing the problem with plotting vr is related to your issue about dimensionality above?  The other issue I encountered was several warnings about divide by zeroes and invalid values, e.g. 

analysis/analytical.py:362: RuntimeWarning: divide by zero encountered in divide
/sw/lib/python2.7/site-packages/scipy/interpolate/interpolate.py:520: RuntimeWarning: invalid value encountered in greater
  above_bounds = x_new > self.x[-1]
gandalf/analysis/analytical.py:332: RuntimeWarning: invalid value encountered in double_scalars
  return (2./(self._nDim+2-self._w)) * self.R_s(t) / t

rbooth200:
Yeah, the warnings are associated with trying to plot the solution at t=0, where some of the quantities are undefined. If you're worried about this then it should be possible to run the simulations and make the plots after? This should avoid the problem.

Yeah the issue with v is that |v| isn't defined in 2D, so you can do the comparison, which is why I commented it out. Once we have a way of getting |v|, it would make sense to correct this line.

rbooth200:
I've made some small changes to sedovtest.py and analytical to avoid the error at r=0 by not plotting it. I changed sedovtest.py to not plot until the end to avoid errors at run time. We can also add a small parameter to the time to avoid the warnings.

rbooth200:
Ok, I added a small parameter to the time since it seems to work well.

PR #54, opened 2015-12-03
Solved bug in accretion

giovanni-rosotti:
Silly bug - yet a showstopper...


dhubber:
Yes, definitely a show-stopper!

PR #58, opened 2016-03-29
Added check to kill the simulation when smoothing length is too large…

rbooth200:
… - only if periodic/reflecting boundaries are used.

I've tested it by varying the number of neighbours in a sod shock tube test. It seems to work fine. Not tested with MPI.


rbooth200:
Fixed the compile bug with mpi. Ran a couple more sod tests with mpi, results seem sensible.

giovanni-rosotti:
It seems fine to me. David, do you want to add it also to the meshless?

dhubber:
Yes, I can do this (add to the meshless) before we accept the merge.  I will do it in a few days after my first lecture.

rbooth200:
I've added this to the meshless now since I needed to make a modification the SPH implementation anyway - the condition wasn't yet strict enough to cover all edge cases

PR #59, opened 2016-03-29
Added reflection for velocity when copying mirror ghosts

rbooth200:
Look at the diff - should be straightforward to see why this can be merged.


giovanni-rosotti:
Absolutely - I can't believe this wasn't merged yet

PR #66, opened 2016-09-01
Brute force tree

rbooth200:
I've removed the duplicate brute force code by including a brute force tree instead. This branch also has the current 'best guess' as to what the best time-integration for the meshless should  be


giovanni-rosotti:
Cool, I'm glad that you managed to remove a net of ~2000 lines in this way. I'd like to test it with MPI before merging. Or have you already done it?

rbooth200:
I'm pretty sure that I did some tests, but I don't remember exactly what since it was a little while ago. Feel free to try a couple if you like.

giovanni-rosotti:
Seems fine to me - I've run adshock and freefall with the bruteforce and everything looks sensible...

PR #67, opened 2016-09-01
Fixed error in energy integration with SPH and self-gravity

rbooth200:
This is the missing energy fix

### DIFF
--- a/src/GradhSph/GradhSph.cpp
+++ b/src/GradhSph/GradhSph.cpp
@@ -465,19 +465,18 @@ void GradhSph<ndim, kernelclass>::ComputeSphHydroForces
     if (dvdr < (FLOAT) 0.0) {
 
       winvrho = (FLOAT) 0.25*(wkerni + wkernj)*(parti.invrho + neibpart[j].invrho);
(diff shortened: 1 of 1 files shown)


giovanni-rosotti:
This is fine. Just a quick question before merging: had we really not implemented the change already? Could it be that we have the similar commit in another branch we haven't merged yet?

rbooth200:
I think there might be another fix in papertests, you could use that one instead if you prefer.

giovanni-rosotti:
Do you mean 	008cc7f? Do I understand correctly that the extra 0.5 you added here is actually commented out in that commit, and is added directly only to the dudt of artificial viscosity? Because if yes, when we will merge everything we will get it two times...

rbooth200:
Yep I've just checked and that appears to be correct. Lets just forget this and try to merge papertests asap?

giovanni-rosotti:
ok agreed, let's do as you suggest. but i'll leave the pull request open until i've fixed papertests to remind me that i should action it quickly

rbooth200:
I've tidied up the fix so it's like the one in paper tests and checked that the energy conservation is good in both cases.

PR #68, opened 2016-09-01
Error checking

giovanni-rosotti:
This is trivial enough that I think we should merge it immediately


rbooth200:
Agreed. I will try to resolve the conflicts and merge

rbooth200:
Done, I also made the isPowerOfTwo function inline to supress some warnings.

PR #76, opened 2016-09-16
Slope limiter testing

rbooth200:
I've modified the slope limiters and written some scripts to produce some comparisons between the various slope limiters. Since this branch is built upon the slope_limiter branch we had before I will delete that pull request and we should use this one instead.  This pull request includes the following:

- updating the thermal properties after changes to the conserved quantities
- recomputing the density for particles that become active, only if the density is out of date
- I moved the various particle flags (active, potmin) to be part of the flags class.
- Changes / generalizations to the slope limiter classes
- Tests for the slope limiters.

I've also attached  a couple of results from the slope limiter tests for the discussion. These can be used to rank the limiters by how diffusive they are. I'd suggest the order from least to most diffusive is:

 - Springel 2009
 - Balsara 2004
 - GIZMO
 - TVD Scalar
 - 1st Order Goduonv,

which agrees well with theory.

adsod_rarefaction
adsod_shock
gresho


giovanni-rosotti:
Looks fine to me - I think we could merge, although I'm not really familiar with the meshless and I think David should have the last word. After we have closed this and the tests branch I'll add the slope limiter tests to the tests run after every commit, at the moment we run only SPH tests.

rbooth200:
It might be worth only running a subset of them since they take a few
minutes for each of the gresho tests.

dhubber:
Just looking through some of the code and running through a batch of tests and all looks good so far.  Will run a few more and if they're fine, I'll go ahead and merge.

dhubber:
All fine so I merged the branch.  I guess we can now delete both the slope_limiter_testing and slope_limiters branch?

rbooth200:
Please leave the slope_limiter_testing branch for now. I have some extra 
bits and pieces that I'm still playing around with. I'll either push 
more commits or delete it later...

giovanni-rosotti:
Sorry guys, deleted the branch as soon as I saw the e-mail - feel free to restore it if you need to

rbooth200:
No worries, I'll do that later if need be.

PR #78, opened 2016-09-23
Reallocate

giovanni-rosotti:
This monstruosity should implement the dynamical reallocation. I think this is now working, I had MANY crashes... I don't want do anything related to memory management in the near future!!!
The code seems valgrind clean now.

### DIFF
--- a/src/Hydrodynamics/Hydrodynamics.cpp
+++ b/src/Hydrodynamics/Hydrodynamics.cpp
@@ -162,26 +162,25 @@ void Hydrodynamics<ndim>::CheckXBoundaryGhostParticle
   const FLOAT tghost,                  ///< [in] Expected lifetime of ghost
   const DomainBox<ndim> &simbox)       ///< [in] Simulation domain box
 {
-  Particle<ndim>& part = GetParticlePointer(i);
 
-  if (part.r[0] + min((FLOAT) 0.0, part.v[0]*tghost) <
-      simbox.boxmin[0] + ghost_range*kernrange*part.h) {
+  if (GetParticlePointer(i).r[0] + min((FLOAT) 0.0, GetParticlePointer(i).v[0]*tghost) <
+      simbox.boxmin[0] + ghost_range*kernrange*GetParticlePointer(i).h) {
     if (simbox.boundary_lhs[0] == periodicBoundary) {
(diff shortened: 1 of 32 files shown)


giovanni-rosotti:
Now that this solves #79 and #80 we should be ready to accept it, let me know what you think

src/Hydrodynamics/Hydrodynamics.cpp
```diff
@@ -202,26 +201,25 @@ void Hydrodynamics<ndim>::CheckYBoundaryGhostParticle
   const DomainBox<ndim> &simbox)       ///< [in] Simulation domain box
 {
   if (ndim > 1) {
-    Particle<ndim>& part = GetParticlePointer(i);
 
-    if (part.r[1] + min((FLOAT) 0.0, part.v[1]*tghost) <
-        simbox.boxmin[1] + ghost_range*kernrange*part.h) {
+    if (GetParticlePointer(i).r[1] + min((FLOAT) 0.0, GetParticlePointer(i).v[1]*tghost) <
```
rbooth200: I think this is ugly... Wouldn't it be easier to read if we just copied r and v instead?

src/Tree/Tree.cpp
```diff
@@ -83,11 +83,17 @@ int Tree<ndim,ParticleType,TreeCell>::ComputeActiveParticleList
 //=================================================================================================
 template <int ndim, template<int> class ParticleType, template<int> class TreeCell>
 int Tree<ndim,ParticleType,TreeCell>::ComputeActiveCellList
- (TreeCell<ndim> *celllist)            ///< Array containing copies of cells with active ptcls
+ (TreeCell<ndim>** celllistaux)            ///< Array containing copies of cells with active ptcls
```
rbooth200: I'm not keen on a function that allocated memory and leaves someone else to clean up after it.  It's begging for someone to create a memory leak. Can we not instead use vectors here?

giovanni-rosotti: You've got a point, see 	06ce03c

src/Hydrodynamics/Hydrodynamics.cpp
giovanni-rosotti: Fixed, see de177c2

PR #81, opened 2016-09-26
Iorig

giovanni-rosotti:
Added the ability in python to access iorig. A bit of hacking as it is technically stored as a float in C++, but I actually the copy the memory from the original integer. So if interpreted as a float it just contains garbage; python on the other side reinterprets it correctly.

### DIFF
--- a/analysis/data_fetcher.py
+++ b/analysis/data_fetcher.py
@@ -22,7 +22,7 @@ import numpy as np
 from swig_generated.SphSim import UnitInfo
 from facade import SimBuffer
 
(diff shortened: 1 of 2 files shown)


dhubber:
Oh, why is it held as a float and not as an integer?  I'm confused :-S

giovanni-rosotti:
Because I am too lazy to have C++ support arbitrary types when iorig is the only non-float quantity we will ever use. It would be a lot of work, especially to make it interoperable with Python, and I think it's not worth it

PR #83, opened 2016-09-27
Levelneib

rbooth200:
Added levelneib calculation to meshless.

### DIFF
--- a/src/MeshlessFV/MeshlessFVTree.cpp
+++ b/src/MeshlessFV/MeshlessFVTree.cpp
@@ -491,6 +491,8 @@ void MeshlessFVTree<ndim,ParticleType,TreeCell>::UpdateGradientMatrices
         // Make local copy of hmask for active particle
         hmask = mfv->types[activepart[j].ptype].hmask;
 
+        activepart[j].levelneib = 0;
+
         for (k=0; k<ndim; k++) rp[k] = activepart[j].r[k];
         hrangesqdi = activepart[j].hrangesqd;
         Nhydroaux = 0;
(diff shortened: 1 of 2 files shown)


src/MeshlessFV/MeshlessFVTree.cpp
```diff
@@ -491,6 +491,8 @@ void MeshlessFVTree<ndim,ParticleType,TreeCell>::UpdateGradientMatrices
         // Make local copy of hmask for active particle
         hmask = mfv->types[activepart[j].ptype].hmask;
 
+        activepart[j].levelneib = 0;
+
```
giovanni-rosotti: This doesn't initialise levelneib for inactive particles. Is it correct?

rbooth200: It's what is done in the SPH code too, but I think it is correct. The code does the following:
1) Calculates levelneib for active particles
2) Increases levelneib for inactive particles, if any of the newly active particles have a higher level (smaller time step) than the inactive particles had before.

We don't need to worry about the case that levelneib actually decreases for inactive particles, although it might allow us to increase their timestep if we did. But I'm not aware of anyone doing that since the main purpose is to detect and reduce the timestep when in the region of strong shocks.

giovanni-rosotti: Yes you're right, I made confusion between this and the zero-ing of the levelneib array before the beginning of the loop over cells. Btw, I realise that that is done only up to Nhydro; surely this is wrong and should be up to Ntot...

giovanni-rosotti: After merge, yes. You haven't pushed the commit yet though. I guess technically we don't have to loop over periodic and mpi ghosts, but the difference is so small that I don't care

rbooth200: Oops. Done

PR #84, opened 2016-09-27
Mfvoptimization

rbooth200:
This branch contains a bunch of fixes and optimizations that i have made to the meshless. I'm sure there will be more to add yet, but I think it is a good time to add these to the master since the meshless appears to be working well in pure hydrotests since we added the levelneib calculation...

Next I plan to remove a few more unncessary variables from the particle structure and fix what I think might be a bug in the dust related to individual particle timesteps.


Makefile
```diff
@@ -20,11 +20,11 @@
 #==================================================================================================
 
 
-CPP                =
+CPP                = 
 PYTHON             = python
 COMPILER_MODE      = DEBUG
 PRECISION          = DOUBLE
-OPENMP             = 1
+OPENMP             = 0
```
dhubber: Yes, I agree.  Also, shouldn't the default for COMPILER_MODE be FAST rather than DEBUG since that'll be the default/most common mode when people want to download and use the code?  (DEBUG hopefully only being used rarely for obvious reasons).

src/MeshlessFV/MfvCommon.cpp
```diff
@@ -45,19 +45,39 @@ using namespace std;
...
+
+  // Local references to parameter variables for brevity
+  map<string, string> &stringparams = params->stringparams;
+
+
+  // Riemann solver object
+  //-----------------------------------------------------------------------------------------------
+  string riemann_solver = stringparams["riemann_solver"];
+  if (riemann_solver == "exact") {
+    RiemannSolverType = 1 ;
```
dhubber: Yes, this would probably make this more readable if we left it as it is.  However, I was hoping at some point in the near future we could have a more adaptive Riemann solver, i.e. one which uses the linear/HLLC in most cases but only uses the exact solver in some cases where the approximate solvers break down.  So, if we do that soon, then there's no need to tidy this up too much.

src/MeshlessFV/MfvMusclSimulation.cpp
```diff
@@ -148,12 +143,13 @@ void MfvMusclSimulation<ndim>::MainLoop(void)
 
 
 
-  // Calculate terms due to self-gravity
-  if (mfv->self_gravity == 1) {
+  // Calculate terms due to self-gravity / stars
+  if (mfv->self_gravity == 1 || nbody->Nnbody > 0) {
```
dhubber: Yes, we could do this.  I guess the problem is the SPH forces are all computed in the same routine whereas the meshless is effectively split between hydro and gravity.  So, perhaps we should write a 'ComputeStarGravForces' routine common to all hydro routines rather than bury the in the NeigbhourSearch routines that we currently do?

dhubber:
So, I didn't fully run with this new branch yet because I wanted to find the bug with the meshless parallel block timesteps.  However, I have since convinced myself that the bug actually is not in the block timesteps, due to the fact it happens with global timesteps also.  It does still happen with parallel switched on, so is probably a parallelisation bug hidden somewhere.  I'll keep hunting this down in the meantime (I might create an issue for this separately since it's clearly a separate bug which is persisting).

Makefile
rbooth200: I agree that we should have some defaults and stick to them, but I'd say that I don't know what they are since we've been changing them all the time...

I took this opportunity to make a sorely needed change, with is to make it possible to specify MPI=1 in the main makefile so that those people who have the mpi library installed, but not the compiler wrappers can still compile with mpi easily. Since this already happens on cosmos, it's bound to happen elsewhere also.

src/MeshlessFV/MfvCommon.cpp
rbooth200: Using enums now, You should see that is still clear here, and is clearer in MfvMuscl.cpp

src/MeshlessFV/MfvCommon.cpp
```diff
@@ -259,142 +288,59 @@ void MfvCommon<ndim, kernelclass>::ComputePsiFactors
...
 
-  for (k=0; k<ndim; k++) part.vreg[k] *= part.invh*part.sound;  //pow(part.invh, ndim);
 
+  // Invert the matrix (depending on dimensionality)
+  InvertMatrix<ndim>(E,part.B) ;
 
+  // Complete the calculation of the gradients:
+  for (var=0; var<nvar; var++) {
+    for (k=0; k<ndim; k++) {
+      part.grad[var][k] = DotProduct(part.B[k], grad_tmp[var], ndim) ;
```
rbooth200: It works by changing the order of the sum. By avoiding all the extra kernel evaluations it can save a few per cent at least.

src/MeshlessFV/MfvMusclSimulation.cpp
rbooth200: I added a swicth to the self-gravity function so that if there are stars present, but no self-gravity the function computes the star forces for the active particles, but not the self-gravity, so we do always get the force from the stars now. The reason I put the force from the stars in this function is that its called at the end of the timestep, whereas the hydro is done at the start of the step in the meshless.  We could instead have a seperate function to do the stars, but that would involve an extra loop over active particles so I thought is solution was best.

rbooth200:
> However, I have since convinced myself that the bug actually is not in the block timesteps, due to the fact it happens with global timesteps also.

Block time-steps in the meshless are still far from ideal though since levelneib is calculated using the time-steps from the old set of levels.

Makefile
giovanni-rosotti: I think OpenMP should be switched on by default. MPI is useful only if you are running on multiple machines, so it makes sense is not there by default. But really I can't see a reason why you wouldn't want openMP. Fixed this in 33c1ec6

giovanni-rosotti: Also, if I don't change the compiler (e.g. leaving it blank) and I just switch on MPI, will it work? On my machine compilation succeeds (the compiler can find the MPI headers) but linking does not (the linker cannot find the MPI library). I guess what is going to happen depends on the specific setup, but it might be confusing that one switches on MPI and actually it doesn't work. What do you think? I propose the following:
a) if CPP is set to mpic++ (or variations, i.e. mpicpp, mpicxx, mpiCC), MPI is set to 1 automatically. This is needed for people that do have the compiler wrappers so that they have to change one variable in the Makefile rather than two
b) we explain clearly in the userguide that if you switch on MPI and compilation fails, you need to change CPP as well. There is no way of doing this automatically as we can't know the right compiler to use on the user's machine...

src/MeshlessFV/MfvMusclSimulation.cpp
giovanni-rosotti: So I realised I said something stupid, so forget entirely about this comment. The force of the stars onto SPH particles is always included, even when self-gravity is not included. This is because also in UpdateAllSphHydroForces we call ComputeStarGravForces

Makefile
rbooth200: Ok, I'd forgotten to add the libmpi++ which we seem to need. 


I would suggest that we discourage people from setting MPI=1 themselves since the mpic++ compiler does a much better job of finding the necessary headers/libraries. For example on my laptop I need to specify the location of the MPI header. I'd suggest that by default we don't even list the MPI=0 line, but state in the documentation how to go about doing it if you don't have the MPI wrappers. I.e.:
1) set MPI=1 in the makefile
2) If compilation fails you may need to:
 a) set INCLUDE to point to the directory containing mpi.h
 b) modify the link options, -lmpi -lmpi++ to reflect the library installed.

dhubber:
I just pulled the latest master branch (that now has this pull request merged) and tried to compile on my Mac and I got these python/swig related errors

swig -c++ -python -outdir ../analysis/swig_generated -DGANDALF_DOUBLE_PRECISION -DDEBUG1  -I-I/Users/dah/code/gandalf/src/Headers Simulation.i
Simulation.i:194: Error: Unable to find 'numpy.i'
Simulation.i:219: Error: Unable to find 'HeaderInfo.h'
Simulation.i:221: Error: Unable to find 'Precision.h'
Simulation.i:222: Error: Unable to find 'RiemannSolver.h'
Simulation.i:223: Error: Unable to find 'Simulation.h'
Simulation.i:224: Error: Unable to find 'Parameters.h'
Simulation.i:225: Error: Unable to find 'SimUnits.h'
Simulation.i:226: Error: Unable to find 'Sph.h'
Simulation.i:227: Error: Unable to find 'SphSnapshot.h'
Simulation.i:228: Error: Unable to find 'Render.h'
Simulation.i:229: Error: Unable to find 'SmoothingKernel.h'
Simulation.i:230: Error: Unable to find 'UnitInfo.h'
Simulation.i:231: Error: Unable to find 'CodeTiming.h'
make[1]: *** [Makefile:229: Simulation_wrap.cxx] Error 1
make[1]: Leaving directory '/Users/dah/code/gandalf/src'
make: *** [Makefile:63: gandalf] Error 2

Maybe it's just a Mac-related issue, but any ideas what changes could have caused it?  I couldn't see any changes to the Simulation.i file so maybe it's elsewhere (maybe in one of the Makefiles?).  Btw, the executable compiles and runs fine so no issue there, just the python.

PR #89, opened 2016-10-11
Timing

rbooth200:
This branch contains a huge number of changes, which I now believe to be working. They fall into two categorties
 1) Refactoring to remove the TreeCell template dependency
 2) Refactoring CodeTiming to make it work with MPI and safer to use in general.

You'll see that there are a few other things ended up in here along the way unfortunately.


src/Mpi/MpiKDTreeDecomposition.cpp
```diff
@@ -402,8 +402,8 @@ void MpiKDTreeDecomposition<ndim, ParticleType >::LoadBalancing
       int icell = mpitree->g2c[inode];
 
       // Create bounding boxes containing particles in each sub-tree
-      for (k=0; k<ndim; k++) mpinode[inode].domain.boxmin[k] = mpitree->tree[icell].boxmin[k];
-      for (k=0; k<ndim; k++) mpinode[inode].domain.boxmax[k] = mpitree->tree[icell].boxmax[k];
+      for (k=0; k<ndim; k++) mpinode[inode].domain.min[k] = mpitree->tree[icell].boxmin[k];
+      for (k=0; k<ndim; k++) mpinode[inode].domain.max[k] = mpitree->tree[icell].boxmax[k];
```
giovanni-rosotti: Is this right that in mpitree the variables are still called boxmin and boxmax? I thought you had changed TreeCell to have a variable called bb inside.

src/Tree/BruteForceTree.cpp
```diff
@@ -461,25 +462,33 @@ void BruteForceTree<ndim,ParticleType,TreeCell>::StockTreeProperties
...
+
+    // Else walk through linked list to obtain list and number of active ptcls.
+    while (i != -1) {
+      if (i < Ntot && partdata[i].flags.check_flag(active) && !partdata[i].flags.is_dead())
+        celldata[c].Nactive++;
+      if (i == ilast) break;
+      i = inext[i];
+    };
+
+  }
```
giovanni-rosotti: Wait a second. Was the old version only updating Nactive in the first cell?! This new version instead is updating only the leaf cells. Is this what we want?

src/Tree/HydroTree.cpp
```diff
@@ -1153,7 +897,7 @@ void HydroTree<ndim,ParticleType,TreeCell>::UpdateHydroExportList
           {
             index_cell = Ncellexport[j]++;
           }
-          cellexportlist[j][index_cell] = cellptr;
+          cellexportlist[j][index_cell] = cell.id ;
```
giovanni-rosotti: Is id simply the position in the array? This is a generic question rather than about this specific code, but I am just wondering, why do we need to store it?

src/Tree/Tree.cpp
```diff
@@ -1457,6 +1626,235 @@ FLOAT Tree<ndim,ParticleType,TreeCell>::ComputeWorkInBox
...
+   int cactive = celllist.size();
+   int Nactivetot=0;
+   for (int cc=0; cc<cactive; cc++) Nactivetot += celllist[cc].Nactive;
+   for (int cc=0; cc<cactive; cc++) {
+     int c = celllist[cc].id;
+     celldata[c].worktot += twork*(DOUBLE) celldata[c].Nactive / (DOUBLE) Nactivetot;
+   }
+
+   Nactivetot_out =  Nactivetot ;
+}
```
giovanni-rosotti: Why can't we just return an int rather than passing it as an argument?

src/Common/Simulation.cpp
```diff
@@ -188,6 +188,7 @@ SimulationBase::SimulationBase
 #else
   Nthreads = 1;
 #endif
+  timing = new CodeTiming() ;
```
rbooth200: We could do. My reason for doing it this way is that in ProcessParameters etc we tend to copy the timing pointer. So adding it to the destructor meant changing fewer lines of code.

Also, we always interact with the timing object through one of those such pointers, that are always called timing. By using the same name here for the master one then we can always call the timing object in exactly the same way, which sould be less error prone.

src/Makefile
```diff
@@ -57,7 +57,7 @@ GCCOPT   += -O3 -ffast-math -fPIC -fno-exceptions -fno-rtti
 ICCOPT   += -O3 -fast -fPIC -fno-exceptions
 CLANGOPT += -O3 -ffast-math -fPIC
 else ifeq ($(COMPILER_MODE),DEBUG)
-GCCOPT   += -O3 -g -Wall -Wno-unknown-pragmas -Wno-reorder -fbounds-check -fPIC  #-fprofile-arcs -ftest-coverage
+GCCOPT   += -O3 -g -Wall -Wno-unknown-pragmas -Wno-reorder -fbounds-check -fPIC  -fno-rtti #-fprofile-arcs -ftest-coverage
```
rbooth200: This was because I was trying to use dynamic_cast when testing and I hadn't noticed that the reason that it was compiling was because the -no-rtti option was missing. Thus the reasib for  adding it to the debug code is to make sure that same code compiles with both options

src/MeshlessFV/MeshlessFV.cpp
```diff
@@ -565,7 +562,7 @@ void MeshlessFV<3>::InitialSmoothingLengthGuess(void)
   for (i=0; i<Nhydro; i++) {
     MeshlessFVParticle<3>& part = GetMeshlessFVParticlePointer(i);
     part.h         = h_guess;
-    part.hrangesqd = kernfacsqd*kernp->kernrangesqd*part.h*part.h;
+    part.hrangesqd = kernp->kernrangesqd*part.h*part.h;
   }
```
rbooth200: I think I removed this everywhere. Let me know if I missed some places.

src/Mpi/MpiKDTreeDecomposition.cpp
rbooth200: I had, but the cells of mpitree are not normal tree cells. I can change them if you prefer.

src/Common/Simulation.cpp
giovanni-rosotti: Well, ok, I accept this requires fewer modications the code - let's leave it like this

src/MeshlessFV/MeshlessFV.cpp
giovanni-rosotti: Sorry you are right - I don't know why I got confused

src/Tree/BruteForceTree.cpp
rbooth200: We only need the active particle list for the leaf cells since those are what is returned by compute active cell list, so I believe this now correct. I don't know why it was working before.

src/Mpi/MpiKDTreeDecomposition.cpp
giovanni-rosotti: Yes I think it would be more consistent with the rest of the code. I think MpiTree uselessly duplicates KDTree, but this is clearly outside the scope of this pull request

src/Tree/HydroTree.cpp
rbooth200: It just the position in the array. At the moment it is useful, so that is perhaps a good reason to store it. We could change how we work with cells to avoid the problem though.

src/Tree/Tree.cpp
rbooth200: Could do, but this avoided some warnings when the value was set but not used in the density/force routines. It's only needed when OUTPUT_ALL is defined after all.

src/Tree/BruteForceTree.cpp
giovanni-rosotti: Well I think this is only needed when you don't reconstruct the tree at every step, so probably that's why we never spotted the problem. In any case yes, now it's correct

src/Tree/Tree.cpp
giovanni-rosotti: ok

dhubber:
ok, just had a quick review of that branch.  Most of the changes seem fine, in particular the timing, box and gravity tree changes although there's a hell of a lot of code changes so couldn't check everything.  Some parts I wasn't completely sure what was going on (such as the append/unpack bytes routines) but I'll trust that it is a necessary improvement.  Maybe the only thing I will say is that some of these new functions/classes need some comments (like in the standard class or function header styles we've tried to use) since in some cases it is not (immediately) clear to me its purpose otherwise.  I will have a look at some other parts now and run a few tests just to check it compiles and runs fine on my machine.

rbooth200:
I'll try to find and add the missing class/function descriptions tomorrow. The add bytes/unpack stuff is a horrible hack, but its basically unavoidable and I did it to reduce the need for book keeping as much as possible.

 Otherwise the code in hydro tree was about moving tree walks into the tree and splitting them up from the neighbour search side of things

rbooth200:
Ok, I think I've added all the comments. Can you let me know if there is anything else? Otherwise we were going to hold off the merge til after the mpi_meshless.

rbooth200:
This branch should now be good to go...

PR #91, opened 2016-10-14
Timestep limiter

rbooth200:
I'm beginning this pull request to spark discussion, rather than because it is ready to merge. It contains two methods to limit the timesteps:
- 1) A fully conservative and predicive one based on a tree walk 
- 2) A non-conservative Saitoh & Makino type limiter adjusting the timestep in a post-hoc way.

The first one works by finding the largest dt such that  r_ij \* vsig_ij > dt for all pair in the simulation. This is done in an extra tree walk, which will need making MPI compatible later.

The second one just checks for particles that have too big timesteps compared to their neighbours and reduces their timestep at the next synchronization point. It is not exactly conservative since dQdt is used to predict dQ, but it seems to work fine for the sedov test.

In doing this work, I noticed that we don't enforce the levelneib heirachy properly at the moment. While the issue of reducing the time-step particles in the middle of their step conservatively is near impossible (although limiter 2 does is this in a non-conservative way). However, in addition we don't enforce the hierarchy at the start of the timestep either, which I think is a significant flaw. The reason that the heirachy is not enforced is that levelneib is computed before the new timesteps are evaluated, and after the new timesteps are calculated the levels shift, leading to some particles needing a smaller timestep than the one they are on because some of their neighbours are now on smaller timesteps.

I  think that enforcing the hierarchy would improve the results when no extra timestep limiters are used or when the non-conservative limiters are used. This is because the fluxes are computed at the start of the timestep. Thus if we can predict that we are going to need a smaller timestep at the beginning of it we can reduce it immediately allowing a fully conservative update of these particles. With the global limiter switched on fewer particles tend to have level differences greater than requested, but some still do.  

In SPH we enforce the hierarchy by iterating the whole timestep for particles on levels that are too big, but this is difficult to do correctly for the meshless because tracking the update of dQ is difficult. I have included an attemp at this by iterating on only the flux calculation but it doesn't work properly. I believe it would be much easier to iterate on the gradient calculation until the timesteps are set properly, but currently the gradients are done before the timestep calcution. Placing it after the timestep calculation has the issue that levelneib is not known in advance since we calculate it during the gradient calculation. 

I would think that the most effective solution would be as follows: 1) Compute a gather based estimate of levelneib during the density calculation. 2) Compute the new timesteps. 3) Compute the full levelneib during the gradient calculation. To do this we would need to iterate on particles that had their timestep reduced to ensure it is propograted to all its neighbours. Step 3) could also be done in another tree walk to avoid recomputing the gradients.

The questions I have are as follows: How important do you think this is? Should we expend the extra cycles (probably only a few to 10 per cent) on making it more accurate? How should we implement it? I believe before the flux calculation is the only reasonable way.

### DIFF
--- a/src/Tree/HydroTree.cpp
+++ b/src/Tree/HydroTree.cpp
@@ -647,6 +647,57 @@ double HydroTree<ndim,ParticleType>::GetMaximumSmoothingLength() const
   return hmax ;
 }
 
+template <int ndim, template <int> class ParticleType>
+void HydroTree<ndim,ParticleType>::UpdateTimestepsLimitsFromDistantParticles
+(int Nhydro,                                      ///< [in] No. of hydro particles
+ int Ntot,                                        ///< [in] No. of hydro + ghost particles
+ Particle<ndim> *part_gen)                        ///< [inout] Pointer to Hydrodynamics ptcl array
+ {
+  int cactive;                          // No. of active cells
+  vector<TreeCellBase<ndim> > celllist; // List of active tree cells
+  ParticleType<ndim>* partdata = static_cast<ParticleType<ndim>* > (part_gen);
+
+  debug2("[HydroTree::UpdateTimestepsLimitsFromDistantParticles]");
+  CodeTiming::BlockTimer timer = timing->StartNewTimer("HYDRO_DISTANT_TIMESTEPS");
+
+  // Find list of all cells that contain active particles
+  cactive = tree->ComputeActiveCellList(celllist);
+
+#pragma omp parallel default(none) shared(celllist,cactive,cout,partdata)
+  {
+#if defined _OPENMP
+    const int ithread = omp_get_thread_num();
+#else
+    const int ithread = 0;
+#endif
+    int cc;                                    // Aux. cell counter
+    int j;                                     // Aux. particle counter
+    int Nactive;                               // No. of active particles in current cell
+    int *activelist                = activelistbuf[ithread];
+    ParticleType<ndim> *activepart = activepartbuf[ithread];
+
+
+    // Loop over all active cells
+    //=============================================================================================
+#pragma omp for schedule(guided)
+    for (cc=0; cc<cactive; cc++) {
+      TreeCellBase<ndim>& cell = celllist[cc] ;
+
+      // Find list of active particles in current cell
+      Nactive = tree->ComputeActiveParticleList(cell, partdata, activelist);
+      if (Nactive == 0) continue ;
+
+      // Make local copies of active particles & update vsig_max
+      for (j=0; j<Nactive; j++) activepart[j] = partdata[activelist[j]];
+
+      tree->ComputeSignalVelocityFromDistantInteractions(cell, Nactive, activepart, partdata);
+
+      for (j=0; j<Nactive; j++) {
+        partdata[activelist[j]].vsig_max = activepart[j].vsig_max;
+      }
+    }
+  }
+ }
 
 #ifdef MPI_PARALLEL
 //=================================================================================================
@@ -668,6 +719,7 @@ void HydroTree<ndim,ParticleType>::UpdateGravityExportList
 
   debug2("[GradhHydroTree::UpdateGravityExportForces]");
   CodeTiming::BlockTimer timer = timing->StartNewTimer("HYDRO_DISTANT_FORCES");
+
 
 
   // Find list of all cells that contain active particles
--- a/src/Tree/HydroTree.cpp
+++ b/src/Tree/HydroTree.cpp
@@ -649,13 +649,11 @@ double HydroTree<ndim,ParticleType>::GetMaximumSmoothingLength() const
 
 template <int ndim, template <int> class ParticleType>
 void HydroTree<ndim,ParticleType>::UpdateTimestepsLimitsFromDistantParticles
-(int Nhydro,                                      ///< [in] No. of hydro particles
- int Ntot,                                        ///< [in] No. of hydro + ghost particles
- Particle<ndim> *part_gen)                        ///< [inout] Pointer to Hydrodynamics ptcl array
+(Hydrodynamics<ndim>* hydro)                     ///<[inout] Pointer to Hydrodynamics object
  {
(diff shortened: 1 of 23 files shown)


dhubber:
Okay, so first of all, I'm guessing this is implicitly about the meshless scheme since I assume 'conservative' is supposed to be in the context of a conservative FV scheme (and I'm not sure how that would relate to SPH).  In that case, yes the Saitoh & Makino scheme is not conservative.  I've been trying to think of ways to modify it to make it conservative but nothing satisfactory just yet.

Btw, I'm assuming this branch split off from timing since there seems to be plenty of the same code changes there?  In fact, is there any different code since the above discussion talks about ideas and not about what has actually been done?  Just wondering since I was thinking maybe I'm wasting time looking at this same code again :-)  Actually, looking at the commit history there are a few commits more in this branch.  Maybe we should (at some point soon?) accept the timing pull request and rebase this to master so we can see more clearly the code differences before delving too deep into this?

Some immediate questions/clarifications that come to mind
- You suggest a gather-based levelneib computation first and then a 'full' computation in the gradient calculation.  So I assume the 'scatter' element of it is done in the gradient calculation, since obviously one is needed somewhere?
- So the difference here is there will be some do-while loop, similar to that in SphSimulation, but it will iterate over the density/gradient part, near the bottom of the main loop in the meshless?
- I'm assuming this is simply a way of fixing the Saitoh-Makino to work and can work without the need for the Springel-type timestep treewalk?  If we did do a tree-walk, we would also need to include the MPI pruned trees presumably to make sure nothing funny could happen from other nodes.

Assuming I understand it all correctly, my first impression is that it could work sure.  I'll have to think about it some more and look at the ordering of the main loop but seems to make sense.  I have just noticed that the UpdateAllProperties is called twice in the main loop (with self-gravity) but not sure if that makes much difference here.

rbooth200:
Hi David,

This is indeed currently only referring to the Meshless. To see the changes, just compare this with the timing branch. I'll rebase back onto it in a second to make it easer.

I'll try to clarify your questions. At the moment there are two working timestep limiters in this branch. For example try the sedov test with Nlevels=10 and with/without the two limiters.  I agree that the Springel-type tree walk (conservative limiter) will need modifying with MPI. I'll discuss with Giovanni the best way to do this and report back... Obviously we need to get the MPI working with the meshless before we worry about this.

The main other point I was discussing is a bit seperate, and that is we don't enforce the timestep heirachy properly. I was describing what I think is the best way to do this in the meshless because, if we do it, we must do it at the start of the timestep to ensure conservation. At the moment we just accept that its wrong unless the Saitoh & Makino type limiter I added is used, in which case it is done at the end of the time-step once the 'error' in the levels has been detected. 

After discussing it with Giovanni, we thought that we should probably just leave it as it is. With the new Springel-type limiter we don't really need to enforce the timestep hierarchy since the particles should be guaranteed to have small enough steps anyway. While the Saitoh & Makino timestep limiter would benefit from getting the time-step correct at the start of the step, even without with the energy error seems to be very small (1e-6) in test problems such as the sedov test.

giovanni-rosotti:
I should now have implemented MPI with the conservative timestep limiter (note also that maxsound was not initiliased in the KDTree and OctTree, only for the BruteForceTree. I love you valgrind!), it appers to work correctly. The timestep limiter does a remarkable job I should say.

src/Tree/HydroTree.cpp
```diff
@@ -686,11 +710,37 @@ void HydroTree<ndim,ParticleType>::UpdateTimestepsLimitsFromDistantParticles
...
+      for (int j=0; j<Nactive; j++) activepart[j] = partdata[activelist[j]];
 
       tree->ComputeSignalVelocityFromDistantInteractions(cell, Nactive, activepart, partdata);
 
-      for (j=0; j<Nactive; j++) {
+#ifdef MPI_PARALLEL
+      if (!only_imported_particles) {
+		  // Loop over pruned trees as well
+		  // Doing it after the local tree walk is more efficient as vsig_max might have been already increased
+		  // TODO: is it more efficient to have this loop inside or outside the one over particles? In this way we are trashing
```
giovanni-rosotti: I'd like input on this comment. It's probably more efficient to swap the loop over the pruned trees with the loop over active cells (locality of the tree in the cache is far more important than the locality of the cell). Should we bother? It would be a problem for openmp (would you parallelize over trees at this point? it's a waste if there are more openmp threads than mpi jobs). And it's a pattern we are using also in other places, even if it's not optimal.

rbooth200: First off, I think it's great that this is working and I suspect that the possible performance gains here might be small.  However, perhaps it is worth thinking about this more generally.

The code would be easier to understand if we seperated out the look computing the local criterion and the pruned tree walk because we would be able to get rid of a bunch of if(!onlly...) statements making it easier to follow the logic. I'm in favour based on those grounds, along with the fact that I doubt we will see much of a performance difference.

As far as maintaining cache efficiency and thread utilization: would the openMP collapse(2) directive help?

rbooth200: Ok. I'm happy as long as you trim the comment.

giovanni-rosotti:
Done. In general, when we have so little modifications to suggest, we can do them directly.

Are we ready to merge this branch? I think so. David should do it as me and Richard committed code. Let's be quick because the to do list is long.

rbooth200:
I agree that this is ready to go...

dhubber:
ok, having a look at this now.

dhubber:
ok, I've had run a few tests with this, both with and without the timestep limiter (although not with MPI), and it seems to work fine without any problems on my machine.  I also had a quick look at the changes to the code and there's nothing I'm overly worried about.  The only extra thing I'd request before accepting is to add this new 'time_step_limiter' parameter to the userguide.tex file, with a small one sentence description of all options (in the same style as all other parameters).  We probably need to go through and update the userguide at some point anyway but it's probably worth trying to keep new parameters updated whenever they are added.  Then I think it's fine to be merged!

rbooth200:
I've now added the documentation

Issue #62, opened 2016-04-11
Periodic gravity


rbooth200:
There is still a small bug in the periodic gravity routine. This is related to the fact that the nearest periodic neighbour to a cell centre is not necessarily the same as the one closest to a particle within the cell. 

This means that we need to make sure that we have the periodic corrected distance on a particle by particle basis.

rbooth200:
The Ewald correction is also not included when stars are present. It also seems to neglect the quadropole term and might be in error with the fast monopole too.

giovanni-rosotti:
See #73, I think at least now the Ewald correction is present also when there are stars. Is the other staff still a problem or has it been corrected in the meanwhile?

dhubber:
This first issue might be what caused the problems I had with the Jeans test which forced me to effectively go to brute-force.  Is there a simple solution to this that we've simply not done?

And yes, the Ewald correction is now made also with the stars.  I'll need to double check it's handled by all cases though before considering that one solved.

rbooth200:
I think this bug was referring to the edge case where a particle / cell sits nearly half a box away from the cell that we are computing the force for. In this case we can think that dr[k] should be about +0.5 boxsize[k], when really it should be about -0.5 boxsize[k]. To correct for this we need to call NearestPeriodicVector in the gravity tree routines for each particle pair when periodic boundaries are used

giovanni-rosotti:
The issue with the cells too big has been fixed in #115 . Keeping the issue open until we have fixed the Nbody in #114

rbooth200:
What is the status of the ewald correction with the stars now? I think #114 is not going to be merged now since we have most of the commits. 

If possible I'd like to see the ewald correction for the stars being optional in non-selfgravitating simulations. This is useful for cases where periodic boundaries are being used to simulate an infinite medium such as the ISM, but we are interested in a single set of star clusters. The binary accretion test is one case where this would be useful.

dhubber:
> What is the status of the ewald correction with the stars now? I think #114 is not going to be merged now since we have most of the commits.

I think it's working now.  I tried the Jeans test with only stars (so in effect the limit of the Jeans length becoming zero) and it seemed absolutely fine.  #114 has already been closed since as you say it's effectively already been merged into master elsewhere.


[...]

PR #94, opened 2016-10-18
Gravity debug2

dhubber:
This pull request contains various bug fixes, code improvements and other things for the Meshless FV gravity algorithm.  These include : 
- Various minor bug fixes in the general meshless gravity routine
- Correcting the variable smoothing length gravity (i.e. zeta) terms for the meshless
- Adding the analytical solution to the Jeans test
- Simplifying various OpenMP loops (mainly changing private variable to local thread variables)

The original gravity_debug branch got entangled with multiple commits due to rebase conflicts.  So I created a new gravity_debug2 branch and cherry-picked the relevant commits.  All should be there, but I'll double check soon before deleting it completely (from my local machine).

There are still a few issues with this branch which should be sorted before the pull request is accepted.
- Restarting simulations from snapshots does not appear to always work with the Meshless FV with gravity (in the Boss-Bodenheimer test) with various assertion errors being generated.  This suggests either invalid values are being written to snapshots, or particle values are not being correctly converted to the various fluid vectors (e.g. Wprim, Qcons) when reading from snapshots.  I tried also with hydro simulations and it did not appear to be an issue, so maybe this just involves gravitating simulations?
- The Meshless FV scheme will still crash with the masses of some particles dropping to zero.  This may still be a simple fundamental flaw of the meshless scheme rather than an implementation bug.  If that is the case, what should we do about it?  We could throw and exception if self-gravity is enabled without the MFM scheme perhaps?

Once these two issues have been resolved/decided, then hopefully this will be ready for merging.


src/MeshlessFV/MfvMuscl.cpp
```diff
@@ -103,23 +103,21 @@ void MfvMuscl<ndim, kernelclass,SlopeLimiter>::ComputeGodunovFlux
...
+  const FLOAT volume_i = 1.0/part.ndens;
 
   // Loop over all potential neighbours in the list
   //-----------------------------------------------------------------------------------------------
   for (jj=0; jj<Nneib; jj++) {
     j = neiblist[jj];
 
-    FLOAT invh_j   = 1/neibpart[j].h;
-    FLOAT volume_j = 1/neibpart[j].ndens;
+    const FLOAT invh_j   = (FLOAT) 1.0/neibpart[j].h;
```
rbooth200: I hate this. It replaces a single character that is correct with unnecessary and ugly clutter. c++ will automatically cast the into to the smallest floating point type compatible with the others in the calculation.

That means if there are no doubles in the expression, the computation will be done in single precision, otherwise it will be done in double precision. The cast will be done at compile time.  See [here](http://stackoverflow.com/questions/5563000/implicit-type-conversion-rules-in-c-operators). I accept that if the coefficient was 0.5 then the cast is necessary to ensure the calculation is done in single precision.

src/Common/Ewald.cpp
```diff
@@ -512,7 +512,7 @@ void Ewald<ndim>::CalculatePeriodicCorrection
     static double dr1=0.0, dr2=0.0, dr3=0.0;
     static int basemax;
 
-    if (fabs(dr[0]) > 0.55*lx_per || fabs(dr[1]) > 0.55*ly_per || fabs(dr[2]) > 0.55*lz_per) {
+    if (fabs(dr[0]) > 0.6*lx_per || fabs(dr[1]) > 0.6*ly_per || fabs(dr[2]) > 0.6*lz_per) {
```
rbooth200: We should fix the bug associated with [issue 62](https://github.com/gandalfcode/gandalf/issues/62).

rbooth200:
The compilation error for commit 5388781 is due to a MPI call in an OpenMP region (in block timesteps for the meshless, line 967). Do we do this anywhere else? Usually its a bad idea as it requires a higher level of OpenMP/MPI support. Since were are doing a synchronization here we should probably just exit/re-enter the parallel region.

rbooth200:
I've added a fix in Ewald for the periodic gravity bug. However, I've noticed two things:

1) Ewald gravity requires compilation with GSL. I'm not convinced we check anywhere that the code has been compiled with GSL when using periodic boundaries. 

Here's an off the wall suggestion: We could just ship with the necessary functions as part of gandalf - It's open source GPL code, we are GPL as long as we attribute it properly it should be fine...

2) The Ewald gravity is not included when stars are used. I've not changed this as it would also be needed to be added in the Nbody routines for consistency. I suspect that we should support this, however.

rbooth200:
I said I'd look into what GIZMO is doing for its calculation of rdmdt. As far as I understand it, here is what I have found. 

Like use, GIZMO is using a KDK scheme for the timestep update. However, it is doing the flux calculation at the end of the step, after the drift has been aplied and the gravitational accelerations have been calculated. This flux is then used again for the K at the start of the next step. This way rdmdt is up to date. 

The GIZMO update is equivalent to what we are currently doing in SPH, but not with the Meshless, in which we compute the fluxes at the start of the KDK step. Essentially a single kick for dt/2 is done at the start of the simulation. From then on another dt is done for the double K (KDKKDKKDK) and a single kick for dt/2 is done at the end to finish the simulation.

 We could experiment with a GIZMO style update, it would require an extra flux calculation for dt/2 in the post initial conditions setup, along with moving the density, gradients and flux calculation to the end of the time-step. There is also some subltely in how dQ should be applied doing it this way, especially with block time-steps. 

Let me know your thoughts.

dhubber:
Regarding GSL, yes I think this should be a simple exception thrown in the Ewald constructor if GSL is not compiled with the code.  It currently runs through fine but actually not doing it correctly at all which is not as it should be.  So, I'll add this on my next commit.

I don't think GSL should be shipped also with Gandalf.  It is one of these standard libraries that almost all linux distros will have in their package managers (Mac also) so the way it works now is if you don't provide a path, it simply looks for the headers in the standard places (e.g. /usr/include) and if they are not there, it fails on compilation.  The only addition I would make is perhaps there is some way we can check in the Makefile perhaps if those files are included, and if not, print an error that they cannot be found and cease compilation.  All of this also applies really to the FFTW library.

dhubber:
Changing the Meshless loop to mimic a LFKDK-style update does sound appealing.  Maybe this would also make it easier to use the Runge-Kutta style integration (which is not currently working for gravity) for gravity also, if we decided to fully implement that approach (although we should concentrate just on MUSCL for now).

An extra flux-calculation in the set-up is nothing to worry about.  Re-ordering the loop may cause issues.  We'll have to think carefully, as you mentioned, about how the particles, W and Q variables are updated to make sure we don't screw something up, but should be do-able (obviously if it's already done in such a way in GIZMO).  I'll have to look at the code a little more myself to think about how it should change before doing anything.

dhubber:
Regarding the OpenMP-MPI compilation error, it is also done in SphSimulation.cpp at line 1311 in almost the exactly the same way.  However, I can now see that the SphSimulation uses 'default(shared)' whereas the Meshless uses 'default(none)'.  We've tried to use 'default(none)' everywhere but I guess this was changed (Giovanni?) to get the OpenMP and MPI to work together in harmony here.  Nevertheless, I will change the Meshless to 'default(shared)' and see if it fixes the problem.

rbooth200:
> We'll have to think carefully, as you mentioned, about how the particles, W and Q variables are updated to make sure we don't screw something up, but should be do-able (obviously if it's already done in such a way in GIZMO). I'll have to look at the code a little more myself to think about how it should change before doing anything.


I will have a look into this next week. First with pure hydro to see whether it makes a difference on the gresho test -- I suspect it will. I'm going to start with a global time-step only implementation, which I already know how to do correctly. When I get that working I'll have a think about block timesteps. I've worked out how to do the conservation correctly with block timesteps, the main issue will be how to do the prediction of Qcons/Wprim properly for inactive particles. We should probably skype to chat about this before I get too far.

giovanni-rosotti:
Sorry, just catching up now with the discussion.

1) About GSL, I think in theory we could ship it with GANDALF. The question is: how careful does one need to be in the compilation process? For example I wouldn't want to ship FFT with gandalf, as its optimization depends a lot on the machine and supercomputers typically have their own version (which is far more optimized than the one you would get simply by downloading the library and compiling it). In any case definitely the code that needs GSL should not run if compiled without GSL!!!!

2) I've double checked the level of threading we require in MPI. We require the minimum  MPI_THREAD_FUNNELED (lower than this there is only MPI_THREAD_SINGLE, which means that you can't have multi-threaded code at all), where we promise that we will call MPI functions only from the master thread. That's what we are doing, so it's actually fine. Using default(shared) for this specific case is thus correct (although unpleasant).

3) can we move the discussion to reorganising the main loop to a separate thread? In any case I need to think more about this...

rbooth200:
I don't think GSL does any special optimization. Personally I don't mind having it as a requirement for periodic gravity since it's so common and periodic gravity is not (cosmology aside).  FFT not so much, many people will not make use of it - I don't think I've ever compiled with it enabled...

I'd make sure that the Ewald code only throws if we acutally try to use GSL functions. We often construct the Ewald object and never use it, so I'm not convinced we should throw on construction.

dhubber:
I don't think there's any need at all to ship GSL with GANDALF.  We've already provided a simple approach in the Makefile where the user only has to switch on the GSL (or FFTW for that library) flag to 1 and it will look in the usual/default place (which should cover all the Linux users and most of the Mac users) and if not, then they just add the library paths right there.  I believe that the GSL functions are only used during set-up and therefore any kind of optimisation we could achieve through compiling statically with it would be negligible.  But I should double-check this though!

Regarding throwing an exception, although the Ewald object is often constructed, the tables are only calculated (and therefore used) when at least one of the boundaries is periodic.  So I was suggesting throwing an exception in this case, i.e. one of the boundaries is periodic AND GSL is not included.

dhubber:
I pushed a one-line change with the 'default(shared)' edit but Travis didn't check it for some reason (there was another branch getting checked, but assumed it would be intelligent enough to queue them, but I guess not).  Any way to force Travis to check the branch without actually needing to push another commit??

giovanni-rosotti:
I closed the pull request and reopened it, and this was enough to force Travis (currently rebuilding).

Fine with throwing when the GSL stuff is actually used - please go ahead with it.

rbooth200:
I think there are two things remaining to do on this branch: 

- Add safety checks to the Ewald to make sure GSL is there when its needed (but not complain if its not there but not needed).

- Add Ewald summations for the stars.

I've already corrected the issue with the periodicity of the particles / cells. See commit 2c39d97

giovanni-rosotti:
Unfortunately there are conflicts to be solved before we can merge...

tests/astro_tests/bossbodenheimer.dat
```diff
@@ -7,13 +7,13 @@
 #-----------------------------
 # Initial conditions variables
 #-----------------------------
-Simulation run id string                    : run_id = BB1
-Run SPH simulation                          : sim = sph
+Simulation run id string                    : run_id = BB1-MFV
+Run SPH simulation                          : sim = mfvmuscl
```
giovanni-rosotti: Are we sure we want to change the default scheme in the tests? I don't have a strong opininion about this. At the moment all the tests use sph, which is a bit silly - we obviosly want to test also the meshless. But whatever we do needs to be consistent, we can't have a bunch of tests with the meshless and a bunch with sph.

src/MeshlessFV/MeshlessFVTree.cpp
```diff
@@ -782,18 +764,17 @@ void MeshlessFVTree<ndim,ParticleType,TreeCell>::UpdateGodunovFluxes
       for (int jj=0; jj<Nneib; jj++) {
         i = neibpart[jj].iorig;
         if (!neibpart[jj].flags.is_mirror()) {
-	        if (neibpart[jj].flags.check_flag(active))
```
giovanni-rosotti: Is it correct that we remove "if active"?

src/MeshlessFV/MeshlessFVTree.cpp
```diff
@@ -802,24 +783,21 @@ void MeshlessFVTree<ndim,ParticleType,TreeCell>::UpdateGodunovFluxes
...
     // Could be simply atomic?
-#pragma omp barrier
 #pragma omp critical
     {
       for (i=0; i<Nhydro; i++) {
-	    if (mfvdata[i].flags.check_flag(active))
-	      for (k=0; k<ndim+2; k++) mfvdata[i].dQdt[k] += fluxBuffer[i][k];
-        for (k=0; k<ndim+2; k++) mfvdata[i].dQ[k] += dQBuffer[i][k];
-        for (k=0; k<ndim; k++) mfvdata[i].rdmdt[k] += rdmdtBuffer[i][k];
+        if (mfvdata[i].flags.check_flag(active)) {
```
giovanni-rosotti: This changes what happens to inactive particles. Is it correct?

tests/astro_tests/bossbodenheimer.dat
dhubber: Not particularly.  But what we probably should do is something similar to the paper_tests directory and have separate parameters files for each hydrodynamics algorithm we wish to change, or if not that, then at least rename the parameters file to reflect what algorithm is being used (e.g. bb-gradh.dat).  Any thoughts?

src/MeshlessFV/MeshlessFVTree.cpp
dhubber: I don't think this makes a difference.  The active checks are made else where.  This is simply adding all the contributions together.

src/MeshlessFV/MeshlessFVTree.cpp
dhubber: So here rdmdt and dQdt are only stored at the beginning/end of the timestep (i.e. when active) whereas dQ can receive contributions throughout the step, whether active or not.  I think this is correct now, but maybe Richard should give a comment before deciding that this is 100% fine?

rbooth200: I believe that these change don't actually make a difference. Although we store extra values for dQdt and rdmdt above, we still only use the ones we need here.

I don't really know whether rdmdt is correctly calculated at the moment, I'd have thought that we should build it  up throughout the the timestep. However, I suggest that we forget about it for now - we don't have a working implementation that makes use of it anyway.

tests/gravhydro_tests/jeans.dat
```diff
@@ -0,0 +1,107 @@
+#------------------------------------------------------------------
+# Ewald sine wave test
```
rbooth200: Name is wrong?

PR #102, opened 2016-11-09
Test2

giovanni-rosotti:
See #70 , moved to another branch to cleanup history. I think we are now finally ready to merge this branch! I was able to fix the hybrid test fiddling around with the openMP configuration, so that now all tests run under 10 minutes, which is pretty nice.


rbooth200:
Ok great. Can you explain a  little why the last fix works so well?

giovanni-rosotti:
Because in this way the openMP threads yield when they have finished instead than busy waiting. On an oversubscribed machine, this gives a chance to the other threads to finish their work. Otherwise they need to wait for the OS to schedule them.

rbooth200:
I guess this is only really beneficial in this case of over subscription rather than normal usage.

src/GradhSph/GradhSph.cpp
```diff
@@ -167,9 +167,14 @@ void GradhSph<ndim, kernelclass>::DeleteDeadParticles(void)
...
-  Nhydro -= Ndead;
-  Ntot -= Ndead;
+  // Reduce hydro particle counters once dead particles have been removed and reset all
+  // other particle counters since a ghost and tree rebuild is required.
+  this->NPeriodicGhost = 0;
+  this->Nmpighost      = 0;
+  Nhydro               -= Ndead;
+  Ntot                 = Nhydro;
+
+  // Some sanity checking to ensure there are no dead particles remaining
```
rbooth200: Are we happy with this solution? Is there a case where we might still need the ghosts after deleteing particles? Is there a better solution?

giovanni-rosotti:
> I guess this is only really beneficial in this case of over subscription rather than normal usage.

That's right. But there is no change to the code, only to the travis configuration; no need to worry for the general user

rbooth200:
Can we try adding back the dustbox test to the tests run? It's one that has a simple analytical solution. Or should we delay this for later work?

rbooth200:
Does this [commit](d4e1e25]) fix the problem with Npart=0 for the oct tree? Even if it is broken for another reason we shouldn't let it be broken in even more ways...

src/GradhSph/GradhSph.cpp
giovanni-rosotti: Well in theory the rebuild_tree flag should trigger a search for new ghosts. In practice I don't think it happens in all cases.
I think SPH is fine because we never update the ghosts after my changes in the stock_pruned branch. Whereas the meshless will be wrong at the moment (incidentally, we need to do change here also the corresponding function in the meshless), although we discussed that we might move where we do sink accretion anyway.
I'll do the same change to DeleteDeadParticles for the meshless, but I think it should be fixed in the stock_pruned branch and not here.

giovanni-rosotti:
> Can we try adding back the dustbox test to the tests run? It's one that has a simple analytical solution. Or should we delay this for later work?

It's just a one liner so I will do it. There's still no check for the correctness though.

> Does this commit fix the problem with Npart=0 for the oct tree? Even if it is broken for another reason we shouldn't let it be broken in even more ways...

Which commit? The link doesn't work...

src/GradhSph/GradhSph.cpp
rbooth200: Ok I'm fine with fixing it [there](https://github.com/gandalfcode/gandalf/pull/100)

rbooth200:
I've updated the link...

giovanni-rosotti:
Ok, I've done these two little things. I guess the question about the tree is for David really.

Incidentally, I am not satisfied with having two copies of DeleteDeadParticles as they are exactly the same, but this is another topic...

rbooth200:
I will create an issue for this

giovanni-rosotti:
So my understanding from David is that the relevant changes have been done also the oct tree. Therefore I think this is now ready to be merged.

dhubber:
ok, I just double-checked and I have made the same changes to both the BruteForce and the OctTree as I did to the KdTree.  I also re-run the dustywave test in the virtual machine with the bruteforce and it worked fine.  Since it's still broken, I couldn't obviously test the OctTree but the new additional Npart=0 code segments are visually correct.  I will work on fixing the OctTree in a separate branch, so if that's everything with this one, I think we should be fine to finally merge!

rbooth200:
Ok, I agree that this is ready. merging

PR #104, opened 2016-11-09
Newparticles2

giovanni-rosotti:
See #73 . Doing it in a new branch to cleanup the history. Besides improving comments in a couple of places, I've added the supernova driver call also in the meshless main loop for consistency.
I also got rid of the newParticles flag. The reason is that we already have a rebuild_tree flag, and it was not clear that new flag was introducing anything new. This new way is safer because we can use the existing checks for rebuild_tree.

I believe this is now ready to be merged.

### DIFF
--- a/src/Headers/Hydrodynamics.h
+++ b/src/Headers/Hydrodynamics.h
@@ -84,7 +84,7 @@ public:
   void CheckBoundaryGhostParticle(const int, const int, const FLOAT, const DomainBox<ndim> &);
 
   void CreateBoundaryGhostParticle(const int, const int, const int, const FLOAT, const FLOAT);
-  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int,
+  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int, const int,
                                     const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*);
 
 
--- a/src/Headers/Hydrodynamics.h
+++ b/src/Headers/Hydrodynamics.h
@@ -84,6 +84,9 @@ public:
   void CheckBoundaryGhostParticle(const int, const int, const FLOAT, const DomainBox<ndim> &);
 
   void CreateBoundaryGhostParticle(const int, const int, const int, const FLOAT, const FLOAT);
+  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int,
+                                    const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*);
+
 
 
   // Functions needed to hide some implementation details
@@ -112,6 +115,7 @@ public:
   //-----------------------------------------------------------------------------------------------
   bool allocated;                      ///< Is memory allocated?
   int create_sinks;                    ///< Create new sink particles?
+  bool newParticles;                   ///< Have new ptcls been added? If so, flag to rebuild tree
   int Ngather;                         ///< No. of gather neighbours
   int Nghost;                          ///< No. of ghost particles (total among all kinds of ghosts)
   int NImportedParticles;              ///< No. of imported particles
--- a/src/Headers/Hydrodynamics.h
+++ b/src/Headers/Hydrodynamics.h
@@ -47,6 +47,8 @@ using namespace std;
 template <int ndim>
 class EOS;
 
+class SimulationBase;
+
 static const FLOAT ghost_range = 2.5;
 
 
@@ -85,7 +87,7 @@ public:
 
   void CreateBoundaryGhostParticle(const int, const int, const int, const FLOAT, const FLOAT);
   Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int, const int,
-                                    const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*);
+                                    const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*,SimulationBase*);
 
 
 
@@ -115,7 +117,6 @@ public:
   //-----------------------------------------------------------------------------------------------
   bool allocated;                      ///< Is memory allocated?
   int create_sinks;                    ///< Create new sink particles?
-  bool newParticles;                   ///< Have new ptcls been added? If so, flag to rebuild tree
   int Ngather;                         ///< No. of gather neighbours
   int Nghost;                          ///< No. of ghost particles (total among all kinds of ghosts)
   int NImportedParticles;              ///< No. of imported particles
--- a/src/Headers/Hydrodynamics.h
+++ b/src/Headers/Hydrodynamics.h
@@ -86,8 +86,8 @@ public:
   void CheckBoundaryGhostParticle(const int, const int, const FLOAT, const DomainBox<ndim> &);
 
   void CreateBoundaryGhostParticle(const int, const int, const int, const FLOAT, const FLOAT);
-  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int, const int,
-                                    const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*,SimulationBase*);
+  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const FLOAT,
+                                    const FLOAT, const FLOAT*, const FLOAT*,SimulationBase*);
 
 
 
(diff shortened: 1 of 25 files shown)


giovanni-rosotti:
Sorry, I was forgetting about the last commit. The code now crashes with MPI. I discussed the thing with David and we think that it is premature to implement supernovae with MPI at this stage - things might still change. It's best to wait and do it later in the icsilcc branch.

src/Headers/Hydrodynamics.h
```diff
@@ -84,6 +86,9 @@ class Hydrodynamics
   void CheckBoundaryGhostParticle(const int, const int, const FLOAT, const DomainBox<ndim> &);
 
   void CreateBoundaryGhostParticle(const int, const int, const int, const FLOAT, const FLOAT);
+  Particle<ndim>& CreateNewParticle(const enum ptype, const enum parttype, const int, const int, const int,
+                                    const FLOAT, const FLOAT, const FLOAT, const FLOAT*, const FLOAT*,SimulationBase*);
+
```
rbooth200: Is this pointer to SimulationBase really necessary? I know we use it to flag for a tree rebuild, but I'm not super happy with this solution.

src/Hydrodynamics/Hydrodynamics.cpp
```diff
@@ -123,6 +124,67 @@ Hydrodynamics<ndim>::Hydrodynamics(int hydro_forces_aux, int self_gravity_aux, F
...
+  const int n,                         ///< [in] Current integer time
+  const int level_step,                ///< [in] ..
+  const int level_max,                 ///< [in] Current maximum block timestep level
+  const FLOAT t,                       ///< [in] Current simulation time
+  const FLOAT m,                       ///< [in] Mass of new particle
+  const FLOAT u,                       ///< [in] Specific internal energy of new particle
+  const FLOAT r[ndim],                 ///< [in] Position of new particle
+  const FLOAT v[ndim],                 ///< [in] Velocity of new particle
+  SimulationBase* sim)				   ///< [inout] Simulation object
+{
```
rbooth200: Is it really necessary to have all of this info in the functions interface? It seems to me that this function is trying to do two things: 1) create and 2) initialize the particles. Surely we should default construct the particle and leave 2) to the caller. We are returning a reference anyway and leaving initialization to the caller would be more flexible since e.g. dust particles don't need u, but might need other things setting.

src/Hydrodynamics/SphSimulation.cpp
```diff
@@ -229,6 +229,25 @@ void SphSimulation<ndim>::ProcessParameters(void)
   }
 #endif
 
+
```
rbooth200: I'd prefer a solution more like what I did with the dust: I think that we should set the supernova driver based on a single function call, rather than having to copy/paste this code everytime we need to add it somewhere.

src/Makefile
```diff
@@ -36,7 +36,8 @@ PYFOLDER = ../analysis/swig_generated
 #GTEST = /Users#/david/astro/code/gtest-1.6.0/include
 
 
-VPATH = $(PWD)/src:$(PWD)/src/Common:$(PWD)/src/GodunovSph:$(PWD)/src/GradhSph:$(PWD)/src/Headers
+VPATH = $(PWD)/src:$(PWD)/src/Common:$(PWD)/src/Feedback:$(PWD)/src/GodunovSph
```
rbooth200: Surely we can remove GodunovSph from the path?

src/MeshlessFV/MeshlessFVSimulation.cpp
```diff
@@ -363,6 +363,23 @@ void MeshlessFVSimulation<ndim>::ProcessParameters(void)
     ExceptionHandler::getIstance().raise(message);
   }
 
+  // Supernova feedback
+  //-----------------------------------------------------------------------------------------------
```
rbooth200: See my previous argument

src/MeshlessFV/MfvMusclSimulation.cpp
```diff
@@ -102,6 +102,15 @@ void MfvMusclSimulation<ndim>::MainLoop(void)
   // Advance N-body particle positions
   nbody->AdvanceParticles(n, nbody->Nnbody, t, timestep, nbody->nbodydata);
 
+  // Reset rebuild tree flag in preparation for next timestep
+  rebuild_tree        = false;
```
rbooth200: Why do we set rebuild tree to false here?

src/Nbody/NbodyLeapfrogKDK.cpp
```diff
@@ -97,29 +102,66 @@ void NbodyLeapfrogKDK<ndim, kernelclass>::CalculateDirectSmoothedGravForces
...
+
+    }
+    else {
+
+      // Sum grav. contributions for all other stars (excluding star itself)
+      //-------------------------------------------------------------------------------------------
+      for (j=0; j<N; j++) {
+        if (i == j) continue;
+
+        for (k=0; k<ndim; k++) dr[k] = star[j]->r[k] - star[i]->r[k];
```
rbooth200: I don't like that we are duplicating all this code for the periodic/non-periodic part. Please combine

src/Headers/Hydrodynamics.h
giovanni-rosotti: I don't think there is any other way of doing it that does not force **the caller** to flag for the tree rebuild. And that is much more dangerous than passing this pointer, because who writes the caller will surely forget to set the flag

src/Hydrodynamics/SphSimulation.cpp
giovanni-rosotti: I agree, but this eventually will be solved in my "fixes" branch...

src/MeshlessFV/MfvMusclSimulation.cpp
giovanni-rosotti: Because this is the end of the timestep, and it is just before we inject the new particles. The flag needs to be reset at a certain point (see SPH main loop), and this is the most logical place

src/Makefile
giovanni-rosotti: done in commit 9af6bbf

src/Hydrodynamics/Hydrodynamics.cpp
dhubber: I partly agree with this.  Looking at it now, a lot of this info can simply be taken from the simulation object (using the sim object pointer).  However, the intention of including this other info in the function interface is to force the user to give the minimum required amount of information needed to construct the particle in memory without creating problems (all other quantities important can then be derived from this).  If this wasn't the case, the user might for example only set r and m and forget about u and v and then cause grief elsewhere.  And even if they remembered these four, they might forget (or not even know) to also set r0, v0, etc..  So, I'm in favour of keeping this information in the function interface since I think it's more robust.  I will however remove those other unnecessary things now.

dhubber: One extra thing, this works fine for SPH but we'll need to check that it works for the meshless.  This only sets the particle properties but I haven't thought about calculating the Qcons, Wprim, etc.. fluid vector quantities from this.  Either these conversion functions need to be called in the main loop, or perhaps a Meshless version of CreateNewParticle is required to do everything?  I'm not sure which is the best approach.

dhubber:
ok, I made my changes and pushed them but as you perhaps can see there are conflicts with master.  so I rebased with master and now getting the usual crap where I'm apparently behind the remote version.  So, am I fine in this instance to do a forced-push as Giovanni suggested (git push -f origin newparticles) so as to not mess up the history?

rbooth200:
Its generally necessary to do a forced update after a rebase. Check with gitk first that the commit history looks ok, if so then go ahead...

giovanni-rosotti:
Ok - I can confirm David that the rebase is correct, this branch now stems from master. It's up to Richard though to say whether he is satisfied with the changes.

rbooth200:
Is the periodic correction for the nbody correct? We don't seem to be
updating adot.

With regards to CreateNewParticles I'm not convinced the current solution
enforces correctness either,  since there are many caes in which the
current particle initialization will not work. For example, if we included
chemistry or anything else that requires values to be initialized to a
given range. Given the choice of that function returning something that is
always wrong versus sometimes wrong I think the best solution is always
wrong - then the caller HAS to do think about what they are doing. Allowing
them to not think and create a possibally invalid particle is a really bad
idea and it will bite us in the future.

Ideally we should go further than this: The calling functions that
initialize the particle should only exist for particles that they are valid
for, i.e. the supernova driver should fail if the particle is not SPH
unless e.g. MFV particles are properly supported.

giovanni-rosotti:
Is this still useful now? Or have all these commits gone into final_paper_tests?

dhubber:
I'm pretty sure all the commits have gone into final_paper_tests but I want to check the commits in turn to be sure before deleting.  I'll go through this on the weekend and close it if I'm happy it's unnecessary.

dhubber:
Finally checked through these commits and seems like there's nothing here not already in master (since icsilcc and then final_paper_tests originally came from this branch).  So, I'm going to close this commit.

PR #111, opened 2016-11-30
Dustywave test

rbooth200:
I've added the dusty wave test to the correctness tests, which should be ready to merge with test_correctness branch


giovanni-rosotti:
Fine - I've removed the dust tests from the tests we run without checking the solution as we now have enough tests for the dust already. I'll now merge this

PR #113, opened 2016-12-06
Neighbour manager

giovanni-rosotti:
I started writing the neighbour manager we discussed in the stock_pruned branch. It's currently working only for sph with no self-gravity (and I've temporarily broken everything else). There are a couple of things that I would like to ask wheter it's worth to keep, see the comments I will write shortly.

### DIFF
--- a/src/GradhSph/GradhSphTree.cpp
+++ b/src/GradhSph/GradhSphTree.cpp
@@ -35,6 +35,7 @@
 #include "InlineFuncs.h"
 #include "Particle.h"
 #include "Debug.h"
+#include "NeighbourManager.h"
 #if defined _OPENMP
 #include <omp.h>
 #endif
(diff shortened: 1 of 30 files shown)


src/GradhSph/GradhSphTree.cpp
```diff
@@ -427,46 +393,19 @@ void GradhSphTree<ndim,ParticleType>::UpdateAllSphHydroForces
...
-              sphlist[Nhydroaux] = jj;
-              Nhydroaux++;
-            }
-          }
-          //-----------------------------------------------------------------------------------------
+          const int Nneib=neibmanager.GetNeibList(activepart[j],hydromask,&sphlist,&neibpart,&drmag,&dr,levelneib);
 
           // Compute all neighbour contributions to hydro forces
-          sph->ComputeSphHydroForces(i,Nhydroaux,sphlist,drmag,invdrmag,dr,activepart[j],neibpart);
+          sph->ComputeSphHydroForces(i,Nneib,sphlist,drmag,NULL,dr,activepart[j],neibpart);
```
giovanni-rosotti: Do we really need to pass in the vectors drmag and dr to ComputeSphHydroForces? I can see why we don't want to recompute them. But the meshless for example as far as I can see doesn't use them (although, they are passed in! I guess this comes from copying and pasting).
Maybe the best thing to do is to make neibpart of a different type than the whole particle. And dr and drmag can be saved inside there, so that we don't need these supporting arrays which are annoying.

rbooth200: But I support the idea of using an extended / reduced neighbour particle type.

giovanni-rosotti: They have been removed from the calculation of the flux, but they are still there for the gradients (not to mention h, but that is a separate topic given that it doesn't use neighbours on the fly). They don't seem to be used though. In any case this tells me that I need to rebase onto master...

PR #114, opened 2016-12-06
Icsilcc

dhubber:
Although this branch still needs some revision and testing before merging, I've made this pull request as a place to comment and discuss anything on the branch.  Contains : 
- A total refactoring of the initial conditions class, with each individual IC in its own class.  This is principally to make it easier to use regularisation/relaxation algorithms to generate more complex initial conditions (i.e. which are not simply uniform or spherically symmetric)
- Ewald gravity added to all N-body classes.  Not fully tested (suspect I've forgotten to add code in one or two places)  but most of the work is done.
- Added some new initial conditions, such as ICs for SILCC like simulations and filaments
- Added random supernova driver, for testing, and code for SILCC supernova driver for future


src/Common/SimulationIC.hpp
```diff
@@ -82,106 +83,98 @@ void Simulation<ndim>::GenerateIC(void)
   }
 
 
-  // If not a restart, generate initial conditions either from external file or created on the fly.
+  // Gnerate initial conditions either from external file
```
giovanni-rosotti: The comment is not quite right...

src/Common/SimulationIC.hpp
```diff
@@ -190,11 +183,15 @@ void Simulation<ndim>::GenerateIC(void)
...
+
+  // Finally, generate the initial conditions
+  icGenerator->Generate();
+
   // Scale particle data to dimensionless code units if required
   if (rescale_particle_data) ConvertToCodeUnits();
 
   // Check that the initial conditions are valid
-  icGenerator.CheckInitialConditions();
+  icGenerator->CheckInitialConditions();
```
giovanni-rosotti: I don't like too much carrying around the pointer to the initial condition object for the whole duration of the simulation to be honest. Is it needed only in RegulariseParticleDistribution? If yes, can't we find another way of letting the pointer get there?

src/GradhSph/GradhSph.cpp
```diff
@@ -261,7 +261,7 @@ int GradhSph<ndim, kernelclass>::ComputeH
 
     // Density must at least equal its self-contribution
     // (failure could indicate neighbour list problem)
-    assert(parti.rho >= parti.m*parti.hfactor*kern.w0(0.0));
+    assert(parti.rho >= (FLOAT) 0.999*parti.m*parti.hfactor*kern.w0(0.0));
```
giovanni-rosotti: There's no need for the 0.999 factor. Also, the assert should be using the w0_s2 kernel rather than w0. I think I have already corrected it already in some other branch...

src/Headers/DomainBox.h
```diff
@@ -75,12 +75,12 @@ MPI_Datatype CreateBoxType (Box<ndim> dummy) {
 /// \date   03/04/2013
 //=================================================================================================
 template <int ndim>
-struct DomainBox : public Box<3> {
-  boundaryEnum boundary_lhs[3];     ///< LHS boundary types
-  boundaryEnum boundary_rhs[3];     ///< RHS boundary types
-  FLOAT size[3];                    ///< Side-lengths of bounding box
-  FLOAT half[3];                    ///< Half side-lengths of bounding box
-  bool PeriodicGravity ;               ///< Whether the domain is using periodic gravity.
```
giovanni-rosotti: The comment said that there was a reason why everything was 3d (but I don't know how important it is). If the comment is not correct anymore for whatever reason, then it should be removed. If instead it is correct that here we hard-code 3d, then we might just as well remove the template from the class...

src/Headers/Ic.h
```diff
@@ -0,0 +1,984 @@
...
+    simunits(_sim->simunits), simbox(_sim->simbox),
+    simparams(_sim->simparams), randnumb(_sim->randnumb)
+  {
+  };
+  virtual ~Ic() {};
+
+
+  // Virtual functions
+  //-----------------------------------------------------------------------------------------------
+  virtual void Generate(void) {};
```
giovanni-rosotti: Shoudln't this be pure virtual? Do we need this empty function?

src/Hydrodynamics/SphSimulation.cpp
```diff
@@ -1116,6 +1151,8 @@ void SphSimulation<ndim>::ComputeBlockTimesteps(void)
     level_step = level_max + integration_step - 1;
     dt_max     = timestep*powf(2.0, level_max);
 
+    cout << "LEVEL_MAX : " << level_max << "   " << dt_max << "   " << dt_min_hydro << endl;
```
giovanni-rosotti: This cout should be removed

src/Thermal/PolytropicEOS.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+
+//=================================================================================================
+//  Polytropic::Pressure
+/// Calculates and returns thermal pressure of referenced particle
+//=================================================================================================
+template <int ndim>
+FLOAT Polytropic<ndim>::Pressure(Particle<ndim> &part)
+{
+  return Kpoly*pow(part.rho, eta);
```
giovanni-rosotti: This is the only EOS that does not define pressure as (gamma-1)rho*u. Even the isothermal EOS does that - the reason is that when we call pressure u and rho have already been computed. For this reason, in the neighbour manager branch I actually got rid of the pressure function from the children and only defined it in the parent class. I think the same will happen also to this one

src/Common/SimulationIC.hpp
```diff
@@ -82,106 +83,98 @@ void Simulation<ndim>::GenerateIC(void)
...
-  else if (ic == "ewaldsine" || ic == "jeans" ||
-           ic == "ewaldsine2" || ic == "ewaldslab" || ic == "ewaldcylinder") {
-    icGenerator.EwaldDensity();
+  else if (ic == "ewaldsine" || ic == "ewaldsine2" ||
+           ic == "ewaldslab" ||  ic == "ewaldcylinder" ||
+           ic == "jeans") {
+    icGenerator = new EwaldIc<ndim>(this, hydro, invndim);
   }
-  else if (ic == "gaussianring") {
-	  icGenerator.GaussianRing();
```
giovanni-rosotti: Maybe I'm wrong, but I think the gaussian ring test has disappeared. I think also the blob one. Might be worth to double check that we didn't lose other ones along the way...

src/Ic/TestIc.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <sstream>
+#include "Precision.h"
+#include "Debug.h"
+#include "Ic.h"
+using namespace std;
+
+
+
+//=================================================================================================
+//  TestIc::TestIc
```
giovanni-rosotti: Shouldn't this be called SILCIc or something like that? Test is a rather generic name...

src/Hydrodynamics/SphSimulation.cpp
```diff
@@ -1663,6 +1729,5 @@ void SphSimulation<ndim>::SmoothParticleQuantity
  (const int Npart,
   FLOAT *values)
 {
-
```
giovanni-rosotti: An empty function which is never called has no right to exist. Let's get rid of it!

src/Headers/Simulation.h
```diff
@@ -261,7 +266,7 @@ class Simulation : public SimulationBase
   virtual void ProcessNbodyParameters(void);
   virtual void ProcessParameters(void)=0;
   virtual void RecordDiagnostics(void);
-  virtual void RegulariseParticleDistribution(const int) {};
+  virtual void RegulariseInitialConditions(const int) {};
```
giovanni-rosotti: A grep through the code tells me that this is called only for SPH. We should add it to the meshless

giovanni-rosotti:
I just wrote a series of comments; I haven't tested the nbody yet.

I gave a look at travis, and the failure is coming from the sedov test. Could it be that you have left the supernova switched on?

src/Ic/ContactDiscontinuityIc.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "Precision.h"
+#include "Debug.h"
+#include "Ic.h"
+using namespace std;
+
+
+
+//=================================================================================================
+//  ContactDiscontinuityIc::ContactDiscontinuityIc
+/// Set-up SILCC-type simulation initial conditions.
```
rbooth200: These comments seem wrong...

src/Common/SimulationIC.hpp
dhubber: Yes, one of 1000+ incorrect comments in Gandalf :-)

src/Common/SimulationIC.hpp
dhubber: Yes, I wanted to ask about this.  The test itself has not disappeared (at least not the subroutines that do the work; they're still in Ic.cpp).  But the functions calls (in SimulationIC.hpp) were not there so I never added the objects.  I wasn't sure if they were called somewhere else by mistake or some other issue.  Either way, they're not in IC objects but this can be easily done of course.

src/Common/SimulationIC.hpp
dhubber: Yes, the initial conditions object can in principle be ditched after PostInitialConditionsSetup.  There's still some thinking to do about how to generate the Initial Conditions and the regularisation.  For example, the regularisation was originally used to simply regularise the particles locally (to make them approximately a glass) so their other properties were not assigned.  But if their positions are being drastically changed (as they can be in the new IC regularised iteration) then their properties should be set afterwards (This also ties in to Richard's glass comment later).

src/GradhSph/GradhSph.cpp
dhubber: Yes, not using that kernel is probably what's causing this (will change it next commit or check for the fix elsewhere).

src/Hydrodynamics/SphSimulation.cpp
dhubber: Was once a dream that I didn't have time to actually fulfil.  Not all dreams come true...... Deleted!!

src/Ic/TestIc.cpp
dhubber: Oh, I should probably delete this before we fully accept the pull request.  Just wanted a sandbox class to test things out in without messing up existing code (there's already a SilccIc.cpp elsewhere).

src/Headers/Simulation.h
dhubber: Yes, it should be part of both of them (actually the idea was in part from the Gaburov & Nitadori meshless paper so definitely belongs in there).  Once we've figured out the few remaining issues with the regularisation, then we should be able to transfer it over quite easily.

src/Hydrodynamics/Hydrodynamics.cpp
```diff
@@ -80,16 +81,16 @@ Hydrodynamics<ndim>::Hydrodynamics(int hydro_forces_aux, int self_gravity_aux, F
 
   // Select and construct equation of state object from given parameters
   //-----------------------------------------------------------------------------------------------
-  if ((_gas_eos == "energy_eqn" || _gas_eos == "constant_temp" ||
-       _gas_eos == "isothermal" || _gas_eos == "barotropic" ||
-       _gas_eos == "barotropic2") && gas_radiation == "ionisation") {
+  if ((_gas_eos == "energy_eqn" || _gas_eos == "constant_temp" || _gas_eos == "isothermal" ||
+       _gas_eos == "polytropic" || _gas_eos == "barotropic" || _gas_eos == "barotropic2") &&
+      gas_radiation == "ionisation") {
     eos = new IonisingRadiation<ndim>
```
dhubber: Yes, I agree something better is definitely needed!

src/Headers/Ic.h
```diff
@@ -0,0 +1,984 @@
...
+  virtual FLOAT GetValue(const std::string, const FLOAT *) {return (FLOAT) 0.0;}
```
dhubber: GetValue simply returns the (floating point) value of some quantity (with the given string name) at the given position, for the purpose of defining all IC quantities in one place.  It's hopefully a straightforward function that will cover 99% of ICs that anyone can write.

src/Headers/Ic.h
```diff
@@ -0,0 +1,984 @@
...
+  virtual FLOAT GetDensity(const FLOAT rad) {return (FLOAT) 0.0;}
+
```
dhubber: I'll probably ditch GetDensity since GetValue covers that also.  They are not pure virtual since not all classes are currently designed to generate ICs this way (although that might change as I get round to giving all of them an overhaul).  But that's not a priority atm.

GetValue (the only one that matters since GetDensity is history) does take a ndim position to return the IC value at that location.

src/Common/SimulationIC.hpp
dhubber: Actually, that diff you've highlighted shows I'm talking rubbish since it was there and was deleted!  Hmmmm, maybe I deleted them accidentally and didn't notice.  Anyway, I'll put it back in there now!

src/Headers/DomainBox.h
dhubber: Yes, I was confused why this was hard-coded into 3d in the first place.  The Ewald object is only ever constructed in 3d (and throws an exception if we attempt to use it in 1d/2d) so I'm not sure if this was needed.  If there's a reason to keep it that I've not thought about, we can but otherwise we can get rid of it later.

dhubber:
>I think the reason is that the IC class is trying to do far too much on its own. I feel like the base IC class should be doing just the bare minimum: It should provide some utlity functions for the derived classes such as AddRandomSphere, but no more. I'm not even convinced that functions like AddBinaryStar and GenerateTurbulentVelocityField belong here either.

There are one or two functions that I've moved to other classes (e.g. the Lane-Emden functions are moved to the PolytropeIc class) but some, like the two you mention (AddBinaryStar and GenerateTurbulentVelocityField) are needed by more than one IC class so should (at least for now) live in the IC class.


>I am not convinced that CalculateMassTable should be part of this object and I'm pretty convinced that ContactDiscontinutiy, GaussianRing and SedovBlastWave should not.

If it were to live anywhere, then this is the place it should be (where else would it go without making some other class family?).  However, it's now been deleted anyway since it's not needed now the Monte-Carlo + regularisation ICs are working okay.  

As I mentioned in a previous comment to Giovanni, I've now moved the GaussianRing code to its own class and file.  But I'm not sure what you mean about the other two since they're already part of their own class (unless I forgot some bit of code somewhere?).


>Also, is there code to generate a ndim-dimensional glass in here somewhere?

In a way, but not in an ideal way at all.  Hopefully we can chat about this when we meet soon since it's the last major issue of the IC component of this branch.

dhubber:
So, I've run the two proposed tests for the N-body Ewald gravity.  The first one (a uniform line of stars) works fine and has zero acceleration so that's great.  I also got the Jeans test with stars to work.  However, that reveals a problem since the accelerations seem to be inverted with some weird discontinuity or something.  Basically, I think there's a wrong sign somewhere so will get onto tracking it down.  But once it's found and I can get it to work with this test, then I think that will be that for the N-body Ewald.

dhubber:
ewaldnbody

Results for the Jeans test for the N-body case using the Leapfrog-KDK.  Seems to work pretty well I think so I'm quite happy.  The case I was mentioning earlier was for the Hermite scheme so there is perhaps a problem there that I want to track down first.  Once that's done, and I've cleaned the new code up, I'll push everything later!

giovanni-rosotti:
Tha's great. Can you make this into a test? Presumably it doesn't take long
to run

dhubber:
Yes, I had the same idea to make it into a test (since I'd already done 80% of the work for that).  I've added it to the tests/nbody_tests sub-directory (called nbodyjeans.dat with corresponding python script).  I've still not technically tested this with both gas and stars but the Ewald now works with both separately so that element of this branch is now complete I believe (finally).

giovanni-rosotti:
Cool. So now it's Richard's turn to refactor the IC class. David, I'll turn the python script you've added into a nose test so that it can be run automatically by travis. Is there anything else left in this branch to do?

rbooth200:
Can you point me to where the E-wald correction is done in th N-body force calculation? I remember having some corcerns about it that I'd like to check

rbooth200:
I'm not convinced that we will get the Ewald correction working with the Nbody Hermite integration without a fair chunk of work. The reason is I don't see a way to compute the \dot{a} term from the Ewald correction properly. This needs to be done carefully since some of the terms in the Ewald sum will change sign. We'd need to implement a new function for that...

dhubber:
>  Is there anything else left in this branch to do?

Like I mentioned, it would be nice to test hydro + nbody with Ewald to be sure it all works, but that's not essential right now.  So there's only really Richard's refactoring plus whatever I have to do later to 'fix' it and then it'll be ready.


> Can you point me to where the E-wald correction is done in th N-body force calculation? I remember having some corcerns about it that I'd like to check

Unfortunately, it's done in several places (and I have to admit there's some code repetition that should be eliminated in truth);  every N-body child class in principle can have its own 'CalculateDirectGravForces', 'CalculateDirectSmoothedGravForces', 'CalculateDirectHydroForces' and 'CalculateAllStartupQuantities' which all take simbox and ewald as parameters to compute the Ewald corrections.  

I remember you mentioning about some of them not having 1st time derivative terms computed (with the Ewald correction).  This is partly because I don't know what this term is and if it's possible to determine simply (it might be in the literature but I'm not aware of it).  Luckily it is of less importance due to the 1/r^3 nature of the time derivative but would still be good to do it if possible.

rbooth200:
> I remember you mentioning about some of them not having 1st time derivative terms computed (with the Ewald correction). This is partly because I don't know what this term is and if it's possible to determine simply (it might be in the literature but I'm not aware of it). Luckily it is of less importance due to the 1/r^3 nature of the time derivative but would still be good to do it if possible.

I'm not really willing to accept this solution to be honest. I'd much prefer it if we just disallow Hermite integration with periodic systems.

rbooth200:
So I've made some changes to the particle regularization that should make the interface a bit more flexible. At this stage I've only implemented a bare bones version, with a single example of how it should be used, based on the Silcc ICs. The new implementation means that it *should* now work with the meshless as well as SPH.

David, could you test? Having never used it, I'm not sure what to expect...

dhubber:
Great, thanks.  I'll have a look at it asap!

dhubber:
Richard, a few questions about the regularisation

1) Atm, these changes essentially just encapsulate all of the regularisation functionality in a new class and then this is called immediately after the IC-Generator step (but before the PostInitialConditionsSetup)?  So we still need to add in the step of setting the particle properties (e.g. velocity, internal energy, etc..) after the regularisation step.  But this probably can be done quite easily now immediately after the regularisation itself (inside the if-statement block).

2) For every IC that we wish to use the 'regularizer' (btw, I will un-americanise that later ;-) ), I'm guessing we need to create its own **IC**Regularizer class including the density function?  I'm guessing this also means that currently the GetValue function is not being used (although it or something similar will be needed to set particle properties as discussed in Q1).

3) One of your aims was to be able to generate a glass easily to be used for making ICs.  I'm guessing it should be easy to do here but we should probably make some 'AddGlass' function similar to the other random-particle/lattice functions for ease of use?

4) "ParticleRegularizer<ndim>(simparams, simbox)(hydro, neib, nbody, *reg_func) ;"
What the hell kind of Satanic C++ witchcraft is this?  :-)    It seems like it's maybe calling a constructor but not actually making an object while also calling an operator??

Anyway, in the meantime, assuming I understand it, I'll have a play around checking it's working okay.

giovanni-rosotti:
While you two guys have fun with c++, just an update on why this is crashing on travis. The issue comes from MPI in combination with pure Nbody simulations after I've added the nbody jeans test. I've fixed it in the fix_stars branch, which however is based on neibmanager. I can't replicate the fix as it's in process parameters which was heavily modified, so we would need to modify it again. Shall I skip the offending test in this branch (only when using MPI) for now? Once this has been merged I can then rebase fix_stars onto master and reactivate it there...

dhubber:
Sure, if you're confident you know what the problem is and can fix it later (presumably once we've merged/rebased the various branches into master) then we can skip it for now to get this branch done asap!

giovanni-rosotti:
Cool, I'll do that asap then. Just to warn you, merging this will be a nightmare as there are zillion of conflicts... This branch basically modifies every single line of code in GANDALF!

dhubber:
Merry Christmas to all of us then!  >.<

giovanni-rosotti:
Just to let you know, I've now fixed the tests...

rbooth200:
I've rebased this branch. Wasn't too painful...

dhubber:
Richard, I said a couple of days ago I was happy with the regularisation classes, but on some reflection I have a few (quite big) concerns that I'd like to address now before continuing : 

(a) We're splitting a single IC over two classes (the IC class itself and the regularisation function class) which is perhaps not the best idea.  This is not the end of the world for these simple cases, but if we're doing something more complex like loading in ICs from a file (e.g. from grid/AMR ICs) then it's a problem.  The functions that create (e.g. Generate) and define the ICs (i.e. GetValue, Density, etc..) should really all be in one class, the IC class.
(b) The IC class is one of the few classes, maybe the only class, that we should expect users (the ones who want to get their hands dirty with C++ that is) to create their own versions.  Therefore it should be made as simple as possible and I feel this current way is overly complex.  I don't mind the complexity being hidden away so the (casual) user does not see it or need to bother with it, but in this form they must make their own inherited versions of these various classes, functions, etc.. so they are exposed to it directly.

My basic question/suggestion is can this be simply modified to use the IC object instead of making a brand new RegulariserFunction class?  The GetValue/Density functions will be in the IC class as before and perhaps the IC class can be passed in instead.

Btw, on the subject of how the current regularisation works/performs, I only had to fix one small bug (essentially it goes wrong if h is not initialised to a non-zero value) but otherwise it works fine so there are no other issues on that front.

rbooth200:
Hi David,

Firstly,  we should do away with the global GetValue functions. I genuinely believe that a solution that relies on these cannot be generic, and will cause problems later. Any class that does something useful must not be forced to work with them.


With regards to making life easy for people to write their own IC class, I personally don't mind making them write a second. If they are already writing one derived class, then surely they have the skills to write a second one? 

But I do accept that there will be cases (IcSilcc may be one of them)  where the RegularizerFunction might be better off calling a density function on the IC class. In this case, subclassing DefaultRegularizerFunctionBase to an "ICDefaultRegularizerFunction", which just forwards the density call to an IC->density(part) would work fine. 

One option to make this easy for user to implement is use a pointer to an IC base class, and then call IC->density virtually. However, this isn't good practice since most IC classes don't need a density function, which spreads like a cancer if it's declared in the base the class. So I think a template ICDefaultRegularizerFunction would work nicely, which people could use. 


I will push an example for you to peruse...

Rich

rbooth200:
Ok, Have a look at the changes that I made to DefaultRegularizerFunctionBase ans SilccIc. Hopefully that is simple enough?

dhubber:
> Firstly, we should do away with the global GetValue functions. I genuinely believe that a solution that relies on these cannot be generic, and will cause problems later. Any class that does something useful must not be forced to work with them.

Okay, I still do not understand this argument at all.   What problems exactly?


> With regards to making life easy for people to write their own IC class, I personally don't mind making them write a second. If they are already writing one derived class, then surely they have the skills to write a second one?

Because of the reasons I outlined, this is one case where easy of use must trump over design philosophy.  I do not want us to force users to become C++ gurus in order to use the code, because it is far more likely they will (a) give up, or (b) send us all e-mails about how to do it.  And I don't want either scenario tbh!  This has to be made as simple as possible!

dhubber:
ok, just saw you made another commit as I was posting.  Will have a peek at that now!

rbooth200:
R.e. the 'Ease of Use Argument'. 

I think is a good reason to create an AddUnifromGlass or InsertUniformGlass function. It will have to be slightly different to the AddCubicLattice functions though, since as we discussed it might be near impossible to add multiple glasses this way.

dhubber:
ok, that's definitely an improvement from the previous version from the general users point of view.

Before getting into that too much, I'm now just realising, is your objection to the GetValue function that it can return everything (i.e. rho, v, u, etc..)?  You keep saying 'global' (but not actually explaining your objection fully enough), but is it simply that you'd prefer them to be separated into say GetDensity, GetVelocity, GetInternalEnergy, etc.  functions in the Ic (or whatever) class?

rbooth200:
No, my objection is related to trying to use it to set important particle
quantities. As with the discussion for CreateParticles, this function
doesn't easily allow users to ensure that the particles are valid - they
have no way of knowing what the need to set. While one could come up with a
solution to the latter problem of knowing what needs to be set, it's not
the best approach. There are just too many edge cases that can crop up this
way.

It would be much better to have a ResetParticleProperties function on the
IC class that handles everything. After all, this even supports users
trying to do things we haven't even thought of, since they could do just
about anything they want in that function. This way it's up to the IC
writer to ensure the particle is valid, but then again they are the person
who knows for sure what is valid for their IC.

dhubber:
> No, my objection is related to trying to use it to set important particle quantities. As with the discussion for CreateParticles, this function doesn't easily allow users to ensure that the particles are valid - they have no way of knowing what the need to set. While one could come up with a solution to the latter problem of knowing what needs to be set, it's not the best approach. There are just too many edge cases that can crop up this way.

> It would be much better to have a ResetParticleProperties function on the IC class that handles everything. After all, this even supports users trying to do things we haven't even thought of, since they could do just about anything they want in that function. This way it's up to the IC writer to ensure the particle is valid, but then again they are the person who knows for sure what is valid for their IC.

Okay, I partly see what you're saying now (at long last, lol) in terms of objections to a GetValue function.  But (as you say in the second comment) it's up to the user to decide how to properly construct the ICs and there's no failsafe method that prevents them doing something stupid. And more importantly (as least for my concerns about ease-of-use), this extra layer of regularisation classes seems unnecessarily more complex than having just a (Get)Density function (that is used by the MC and regularisation scheme) and a (Re)SetParticleProperties function in each Ic derived class.  Limiting the Ic class to having just those two functions would seem a much better compromise solution to me than this extra class layer.

Anyway, for now I will have a look at this (Re)SetParticleProperties function idea and see how it works in terms of with and without regularisation.

rbooth200:
So yes, i agree with you that just these two methods on the IC class is the
best way forward.

The reason for the extra layer, as i see it, is that we might want to do
particle regularisation in another part of the code, e.g. when injecting
particles. That's why I wanted to separate the regularisation from the IC
class. The cleanest way to do this is with an extra layer.

dhubber:
ok, still not 100% convinced about the extra class layer (at least how it's done atm; need to think more about it) but it's not super important right this minute.  I just tried adding what we talked about (the 2 functions alone in the Ic class).  Seemed pretty straight-forward so far.  Still need to try it on a better example though (since these ones had zero-velocity ICs with no variations).

The other thing I need to try next is to add the Smoothed Density Ics, since I'll need those pretty soon.  Just need to think how this all clips together in the new classes.

rbooth200:
Looks good.

The only complaint I have at the moment is that the regularization only works with one particle type as it stands. We really want this to be able to simulataneously regularize different density distributions. Can you add that in? I think it means changing the interface to GetDensity, and having the Regularizer function return zero for neighbours that are not included in the density mask.

dhubber:
ok, so I just pushed both a GetDensity with particle types and GetSmoothedDensity options.  I just realised my GetDensity with types is not really done correctly (since it doesn't have the particle type itself) so going to re-do it now!  Definitely worth trying some set of initial conditions (maybe a dust test?) to be sure this is all working fine!

src/Headers/Ic.h
```diff
@@ -1043,42 +1046,118 @@ class RegularizerFunction {
...
   virtual ~DefaultRegularizerFunction() { } ;
 
+
   virtual FLOAT operator()(const Particle<ndim>& part,
                            const Particle<ndim>& neibpart) const
   {
-    FLOAT rhotrue = _rho_func->GetDensity(neibpart.r);
+    FLOAT rhotrue;
+
+    if (regularise_smooth_density) {
```
rbooth200: Can we not just have two different RegularizerFunction function objects? One a DefaultRegularizerFunction and one a SmoothedRegularizerFunction? 

This would be in better agreement with the pricinciples of OOP. I'm happy to do implementation this if desired...

src/Headers/Ic.h
```diff
@@ -104,7 +105,7 @@ class Ic
   virtual void Generate(void) {
     ExceptionHandler::getIstance().raise("Ic::Generate function not implemented");
   };
-  virtual FLOAT GetDensity(const FLOAT *) const {
+  virtual FLOAT GetDensity(const FLOAT *, const Typemask typemask=Typemask(true)) const {
```
rbooth200: I believe this should take a const Typemask&, rather than a const Typemask.

Are we sure that using a Typemask is the right way to go here: It's definitely the most flexible so I won't argue on that ground, but it is a bit of a pain to use when we want just a single particle type. The other option would be to make GetDensity take a ptype. We'd then just call it multiple times if we wanted a 'total density' over some set of particles.

By the way, these functions should not be virtual.  Also, I think they can be declared inline. If so, I'd argue that they should be inline, this would allow the compiler to do much more optimization (particularly when looping over types if we cange the interface).

rbooth200:
As a way to test this, can we do a random cube type test? Instead we'd just use the regularization routines to do the regularization and not take any timesteps (or only 1 perhaps). Afterwards we'd check that the standard deviation of the density is below some threshold.

Dust could be included as a second fluid to regularize in this test and we should check that it works with the meshless, perhaps both the smoothed and unsmoothed varients too?

src/Headers/Ic.h
dhubber: > Are we sure that using a Typemask is the right way to go here: It's definitely the most flexible so I won't argue on that ground, but it is a bit of a pain to use when we want just a single particle type. The other option would be to make GetDensity take a ptype. We'd then just call it multiple times if we wanted a 'total density' over some set of particles.

Yes, this is the main thing I realised after committing.  It's probably best to give it just the particle type (btw, on a slight tangent, but we have both ptype and parttype.  Do we really need both?  Any way to get rid of one because it seems overkill to have two.)  I can make this change now also.

> By the way, these functions should not be virtual.

How can be they not be virtual??  I mean, we are declaring them as pure virtual to force an implementation when they are needed so I'm not sure how else this can be done.  Unless there's some C++ wizardry I'm unaware of!  If they can be inlined then great, since that should speed it up a little.

dhubber:
> As a way to test this, can we do a random cube type test? Instead we'd just use the regularization routines to do the regularization and not take any timesteps (or only 1 perhaps). Afterwards we'd check that the standard deviation of the density is below some threshold.

Okay, I assume you're talking from the point of view of having an automated test here?  That perhaps would be sensible sure.  I'm actually not sure what happens if Nstepsmax = 0.  Perhaps it will only run the Setup and not bother with the MainLoop at all?

src/Headers/Ic.h
rbooth200: > t's probably best to give it just the particle type (btw, on a slight tangent, but we have both ptype and parttype. Do we really need both? 

I think the only place the parttype is used is in the IO routines. I think it only still exists as because we never finished removing it. IIRC I originally had to introduce a second variable ptype to make the dust work. We originally had an itype variable as well, that was used for boundaries etc and was typically set to one of the parttype variables, I managed to get rid of itype eventually, so parttype can probably go now too,

src/Headers/Ic.h
rbooth200: I agree that it would be nice to these functions away from the user, or at least make their life as simple as possible. I'll show you how we can do that once we've got the interface right :)

These really should be two objects though.

rbooth200:
> I'm actually not sure what happens if Nstepsmax = 0.

I don't know either, but Nstepmax = 1 would be fine too.

dhubber:
> Why are we using an int here to decide the type? Surely it should be an enum ptype?

Because I did try to use the enum originally but it was complaining for some reason so switched to int for an easier life.  Actually we were mixing enums and ints before anyway in various places so should stick with one of the two (preferably enums sure, but ints are not end of the world).

rbooth200:
I agree its not the end of the world. Having had a look, the problem is 
because ptype has gone and we only have parttype left.

I'd suggest that we rename parttype -> ptype and make a consciencious 
effort to change int to enum ptype as we go about our business in the 
future?

dhubber:
> I'd suggest that we rename parttype -> ptype

No!!  Part of the issue (at least that I was worried about) was I was using ptype often as a variable (both before and after these changes) and one of the enums was also called ptype.  So, now I've used 'ptype' as the variable name (both in the Particle class and in various subroutines that need it) and the enum is now called parttype.  There's still some messy variable naming I saw in other parts of the code but I tried to get this bit consistent at least.  It's possible some confusion with the naming is what caused my compiling issues that made me use ints.  If we get this all consistent, I'm sure changing the ints back to enums won't be a problem.

rbooth200:
Ok, so we should just conscienciously change int ptype to enum parttype 
ptype...

rbooth200:
When you're happy that this is working as intended, it can go. I'm more or less happy with it as it is.

dhubber:
Closing this pull request (and will soon delete this branch) since all commits are in final_paper_tests (which has additional fixes).

PR #115, opened 2016-12-09
Neibmanager2

giovanni-rosotti:
Combines #100, #107 and #113 (see there for the discussion). I still have to fix a couple of things before we can accept it


giovanni-rosotti:
I think the last commits fix all the problems left in this branch. It should now be ready to be merged I believe.

src/Headers/NeighbourManager.h
```diff
@@ -0,0 +1,329 @@
...
+    directlist.resize(_NCellDirectNeib);
+
+    // Go through the hydro neighbour candidates and check the distance. The ones that are not real neighbours
+    // are demoted to the direct list
+    for (int jj=0; jj < neiblist.size(); jj++) {
+      int ii = neiblist[jj];
+
+      // Compute relative position and distance quantities for pair
+      for (int k=0; k<ndim; k++) draux[k] = neibdata[ii].r[k] - rp[k];
+      GhostFinder.NearestPeriodicVector(draux);
```
rbooth200: Is this correct: At the moment we are not correcting the particles positions like we do in the hydro walk.

giovanni-rosotti: You're right we should do it. But only if it's a hydro neighbour (i.e., using hydromask later); in the gravitating case everything will (or should) be fine anyway thanks to the Ewald correction

rbooth200: Yeah, I think it's only necessary for those ghastly edge cases - like the large cells.

rbooth200:
Assuming this last fix runs ok, I think this branch is now ready...

giovanni-rosotti:
Great to hear. David, it's up to you to say if this is ready given that both me and Richard were involved with it.

dhubber:
Sure, will have a look at it now.

src/Hydrodynamics/SphSimulation.cpp
```diff
@@ -651,18 +545,24 @@ void SphSimulation<ndim>::MainLoop(void)
   nbody->AdvanceParticles(n, nbody->Nnbody, t, timestep, nbody->nbodydata);
 
   // Check all boundary conditions
-  // (DAVID : Move this function to sphint and create an analagous one
-  //  for N-body.  Also, only check this on tree-build steps)
-  if (Nsteps%ntreebuildstep == 0 || rebuild_tree) sphint->CheckBoundaries(simbox,sph);
+  // (DAVID : create an analagous of this function for N-body)
```
dhubber: Yes, I will implement this in the icsilcc branch alongside the N-body Ewald commits.  Luckily the N-body Jeans test was only run for a few steps with no/little boundary crossings.

src/Headers/NeighbourManager.h
```diff
@@ -0,0 +1,321 @@
...
+    *neibdata_p = &neibdata[0];
+
+    return listlength;
+
+  }
+
+private:
+
+  template<class InParticleType>
+   void _EndSearch(const TreeCellBase<ndim> &cell, const InParticleType* partdata, bool keep_direct=true) {
```
dhubber: What exactly does this function do?  Its name suggests it's maybe some last step or deallocation/termination but actually it does a hell of a lot of work and I'm only able to determine that by reading through the whole function.  A lot of these new neighbour classes are perhaps not sufficiently commented, which is fine for most of it since it's quite well-written 'self-commenting' code.  But function names should be clear precisely what they are doing OR have a few sentences explaining what they do in a comment-block (and preferably both).

src/Common/RandomNumber.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+//  GANDALF is distributed in the hope that it will be useful, but
+//  WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+//  General Public License (http://www.gnu.org/licenses) for more details.
+//=================================================================================================
+#include "RandomNumber.h"
+
+
+const FLOAT XorshiftRand::invrandmax = 1.0/1.84467440737095e19;
```
dhubber: Yes, I was hoping there was a different solution to this but if it gets rid of the million or so warnings then so be it!

src/Hydrodynamics/SphSimulation.cpp
rbooth200: This looks important, but I want to get this branch accepted quickly. Are we ok with leaving this for icsilcc?

src/Headers/NeighbourManager.h
rbooth200: I've added a ton of comments...

src/Common/RandomNumber.cpp
rbooth200: I believe this is the 'correct' non c++11 solution. Adding a whole file for one line might seem like overkill, but at least this way interested parties would know where to look.

src/Hydrodynamics/SphSimulation.cpp
dhubber: I'm fine with leaving this for now in order to get that branch finished, but of course we should make an issue for it (as a reminder at least).

Issue #77, opened 2016-09-22
MPI load balancing is still done via brute force


giovanni-rosotti:
After the new domains have been decided, particles need to be re-assigned to the correct domain. This is currently computed by the function FindParticlesToTransfer in HydroTree, called by LoadBalancing in MpiControl. However this function does it via brute-force (i.e. loops over every particle and for each particle loops over all the domains). It seems to me that the function using the tree is already there (FindMpiTransferParticles) in HydroTree, but it is never called. The interface is the same so it really looks like ready to be swapped in. What is going on? Did we just forget to call it, or did it have a fundamental problem?

giovanni-rosotti:
The tree version of the routine looks fine to me and in the branch load_balancing I've switched to it. It's mostly fine, but there is an horrible edge case which breaks it. It happens when there is a particle exactly at a corner between domains, so the particle belongs to different domains. Previously we had a break in the loop over domains, so the particle was going to be sent only to one domain. But now we can no longer do that - how do we deal with this case? Should we set a rule that the particle in this case belongs to, say, the domain on the left? On a related note, how does the tree deal with this?

rbooth200:
I believe it's up to the tree to decide how it deals with these edge cases. They should be very rare in the Oct Tree, but with the KDTree the cell is split based on the number of particles in each cell, which means that if multiple particles fall on the boundary you can end up with some on one side and some the other...

I'm happy with any solution that breaks the degeneracy. Realistically this should be using a criterion like
left < x  <= right, or left <= x < right. I don't care which. 

However, we might have to be careful now with the edges of the simbox. For open boundaries this should be well defined and for reflecting we should hopefully not have particles exactly on the boundary as this might cause problems. For periodic boundaries, I think we need to make sure that the box wrapping is done in exactly the same way

giovanni-rosotti:
I agree, but I think I've stumbled upon a more foundamental problem with the simbox, namely that the MPI domains are not enforced to extend until the end of the simbox (see #117, but I din't realise it affects also periodic boundaries). There is a range between the left edge of the simbox and the leftmost particle which is not owned by any domain... I think the simulation will still run correctly, but clearly we do not want this.

giovanni-rosotti:
[...]

PR #121, opened 2016-12-19
Tree loadbalancing

giovanni-rosotti:
Solves #77

### DIFF
--- a/src/Headers/InlineFuncs.h
+++ b/src/Headers/InlineFuncs.h
@@ -340,43 +340,6 @@ static inline FLOAT clamp (FLOAT value, FLOAT min, FLOAT max)
 
 
 
-
-//=================================================================================================
-//  BoxOverlap
-/// Check if two bounding boxes overlap.  If yes, then returns true.
-//=================================================================================================
-static inline bool BoxOverlap
- (const int ndim,
-  FLOAT *box1min,                      ///< Minimum extent of box 1
-  FLOAT *box1max,                      ///< Maximum extent of box 1
-  FLOAT *box2min,                      ///< Minimum extent of box 2
-  FLOAT *box2max)                      ///< Maximum extent of box 2
-{
-  if (ndim == 1) {
-    if (box1min[0] > box2max[0]) return false;
-    if (box2min[0] > box1max[0]) return false;
-    return true;
-  }
-  else if (ndim == 2) {
-    if (box1min[0] > box2max[0]) return false;
-    if (box2min[0] > box1max[0]) return false;
-    if (box1min[1] > box2max[1]) return false;
-    if (box2min[1] > box1max[1]) return false;
-    return true;
-  }
-  else {
-    if (box1min[0] > box2max[0]) return false;
-    if (box2min[0] > box1max[0]) return false;
-    if (box1min[1] > box2max[1]) return false;
-    if (box2min[1] > box1max[1]) return false;
-    if (box1min[2] > box2max[2]) return false;
-    if (box2min[2] > box1max[2]) return false;
-    return true;
-  }
-}
-
-
-
 //=================================================================================================
 //  BoxOverlap
 /// Check if two bounding boxes overlap.  If yes, then returns true.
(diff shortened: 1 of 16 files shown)


src/Headers/InlineFuncs.h
```diff
@@ -493,10 +456,10 @@ static inline bool ParticleBoxOverlap (SphParticle<ndim>& part, Box<ndim>& box)
 //  ..
 //=================================================================================================
 template <int ndim, template<int> class Particle >
-static inline bool ParticleInBox (const Particle<ndim>& part,const Box<ndim>& box)
+inline bool ParticleInBox (const Particle<ndim>& part,const Box<ndim>& box)
```
giovanni-rosotti: On a different note, why are these functions static? Static means that each translation unit (i.e., cpp file) gets its own version. But we don't care ( a), they will be inlined anyway, and b) if they arent', we are just producing code bloat). In c++, inline is already enough to prevent having problems with multiple definitions... I think all of these functions should just be inline, without static

src/Tree/HydroTree.cpp
```diff
@@ -1449,17 +1449,63 @@ void HydroTree<ndim,ParticleType>::FindMpiTransferParticles
...
-                                        all_particles_to_export.end() - NumPartFound,
-                                        all_particles_to_export.end());
+    all_particles_to_export.insert(all_particles_to_export.end(),
+                                        particles_to_export[inode].begin(),
+                                        particles_to_export[inode].end());
   }
   //-----------------------------------------------------------------------------------------------
 
+#ifndef NDEBUG
+  VerifyUniqueIds(all_particles_to_export.size(),hydro->Nhydro,&all_particles_to_export[0]);
```
giovanni-rosotti: This validation code is quite expensive. It is currently guarded by NDEBUG, so it will not be compiled with DEBUG_LEVEL=0. We have DEBUG_LEVEL=2 (macro VERIFY_ALL), but at the moment the code doesn't even compile with it. Besides, unless we are going to use it on a regular basis (e.g. on Travis), it's useless. What shall we do? Shall I open a separate issue for this?

src/Headers/InlineFuncs.h
rbooth200: I agree.

src/Tree/HydroTree.cpp
rbooth200: We're struggling with Travis with having enough time on travis as it is, so verify all it going to be difficult with travis. But yes, open an issue. We should discuss it at least.

rbooth200: Are we buidling debug or production code on travis?

dhubber: Oh, we definitely do NOT want VERIFY_ALL when using Travis since that will kill us.  Yes, this particular validation is a special brute-force check that we don't want in general.  Maybe it's best to put it in the 'higher' debug flags (i.e. 2).  

Surely we only want production code on Travis?  We only need the debug options (with asserts et al.) when there's actually a problem we want to get to the bottom of!

giovanni-rosotti: Ok, I think this clearly deserves some discussion in its own separate issue...

giovanni-rosotti:
I think this is now ready. It's not passing the tests because it's going overtime with clang, but it's fine with gcc... I'll disable the extra checks and see if we can stay on time. When this is gone, we should think how to reduce the run time on travis because it's painful

rbooth200:
I was looking to acccept this, but it seems there is still a problem with MPI...

giovanni-rosotti:
I swear that when I looked at it this afternoon they were all green ticks! Anyway, I'll give a look at it now...

giovanni-rosotti:
This is now fixed. I've done some cleanup in the last commit, so provided that travis gives green light, from my side it's ready to go. One of the assertions was too stringent; and then there was a problem with load balancing with the meshless that actually was there since a long time. For the record, it had already been fixed in newparticles2. This is why pull requests should be closed as soon as possible, or we are going to fix every problem twice and waste a lot of time in this way.

rbooth200:
I'm happy with this. I'll merge it and close #77

PR #123, opened 2016-12-19
Changed dustybox to 1D

rbooth200:
I've changed the dustybox test to be in 1D.  While I accept that we do need some 3D tests, this one is very slow on travis. It's essentially a 0D problem, so 1D should be fine.


giovanni-rosotti:
Then you need to update also the expected error of the test in nose...

rbooth200:
It's basically the same. I suppose I could add a test for both dust algorithms though...

rbooth200:
If you look at the errors it seems that there is a very small error (10^-9) in the hydro when running with MPI, which is causing the tests to fail. Any ideas?

dhubber:
Is this pull request still needed or have the changes been merged elsewhere (since there have been other dust branches merged)?

rbooth200:
I never Merged these changes because this pull request highlights a likely
small, but un fixed, bug with mpi

Issue #116, opened 2016-12-13
Double precision in python


rbooth200:
We need to be able to use double precision with python for convergence tests, e.g. the sound wave

giovanni-rosotti:
How exactly do we want this to work? Are we happy with just having a compile time flag to select the precision of the python snapshots? Note that this flag will NOT be the same as the precision of the code. The alternative of course would be to be able to select the precision at run time from python. There are only a few situations where one would actually want single precision (i.e. only if there are memory constraints), and therefore increasing the code complexity (it can be done with templates, but they don't play so nicely with SWIG and it would require to add some layers of complexity) to support the flexibility is probably not worth it.

rbooth200:
I'm happy with a compile time flag -  I suspect most users will be too. I also think that this should be seperate from the precision of the code - but I would say that if possible it makes sense to force python to be in single precision if the main code is. I don't think that should be too hard. I don't have a strong opinion on what the default should be - but I'd lean towards double. If users are close to running out of memory are they really going to want to use the python interface anyway?

rbooth200:
Fixed in #124

PR #124, opened 2016-12-19
Doublesnapshot

giovanni-rosotti:
Addresses #116 . Seems to be working fine. The only thing I haven't tested is density_pdf (but I would be surprised if it fails); is it used anywhere? Luckily most things are handled automatically by SWIG. There are only three places where I had to change manually the code. SWIG doesn't handle nicely typedefs so I had to do something manually in the headers, it's ugly but it's not so much code...


Makefile
```diff
@@ -25,6 +25,7 @@ PYTHON             = python
 COMPILER_MODE      = FAST
 PRECISION          = DOUBLE
 OPENMP             = 1
+SNAPPRECISION	   = DOUBLE
```
rbooth200: Can we use another name please, at least in the Makefile. I think that this will cause confusion because people may think it affects the .su format, when it's affecting python. Can we go for PY_PRECISION or PYSNAP_PRECISION or something?

src/Common/Render.cpp
```diff
@@ -321,19 +337,19 @@ int Render<ndim>::CreateSliceRenderingGrid
...
 
-  rendernorm = new float[Ngrid];
-  rgrid = new float[2*Ngrid];
+  rendernorm = new SNAPFLOAT[Ngrid];
+  rgrid = new SNAPFLOAT[2*Ngrid];
 
   // Create grid positions here (need to improve in the future)
-  dx = (xmax - xmin) / (float) ixgrid;
-  dy = (ymax - ymin) / (float) iygrid;
+  dx = (xmax - xmin) / (SNAPFLOAT) ixgrid;
```
rbooth200: We could do away with these casts...

src/Hydrodynamics/RiemannSolver.cpp
```diff
@@ -606,10 +611,10 @@ void ShocktubeSolution::ComputeShocktubeSolution
     riemann->SampleExactSolution(pstar, ustar, s, pl, pr, rhol, rhor, cl, cr, vl, vr, p, d, u);
 
     if (quantity == "x") vals[i] = x;
-    else if (quantity == "rho") vals[i] = (float) d;
-    else if (quantity == "vx") vals[i] = (float)u;
-    else if (quantity == "press") vals[i] = (float)p;
-    else if (quantity == "u") vals[i] = (float) (p/d/(gamma - 1.0));
+    else if (quantity == "rho") vals[i] = (SNAPFLOAT) d;
```
rbooth200: Again, I'm not sure these are necessary...

giovanni-rosotti:
Ok, I've addressed your points - should now be ready to be merged...

Issue #103, opened 2016-11-09
DeleteDeadParticles


rbooth200:
There are currently some things about DeleteDeadParticles that are unsatisfactory:
 
- We have two identical copies of the function
- It invalidates the ghosts and we do not currently have a way to check that the ghosts are valid before using them.

The second problem could be solved pragmatically by just 'flagging' that the ghosts are invalid and checking this flag before walks are done. Alternatively it could be made valid. See the discussion [here](https://github.com/gandalfcode/gandalf/pull/102)

giovanni-rosotti:
The first problem is fixed in #128 . Do we have plans to fix the second one soon, or should we move the issue to a future milestone?

rbooth200:
I will have a think about the second one. There would still be a lot of places in which we need to check the tree's validity. The easiest solution would be to check the flags each (ghost) tree walk. It should be cheap, but its a lot of checks...

rbooth200:
Can we close this issue? DeleteDeadParticles invalidates the whole tree, not just the ghost particles so there is probably not much point in flagging anything.

rbooth200:
Bump

giovanni-rosotti:
I agree it's fine to close this - closing

PR #128, opened 2017-01-23
Delete dead particles

rbooth200:
Removed duplicate copies of DeleteDeadParticles using a template implementation. See #103

### DIFF
--- a/src/Hydrodynamics/RiemannSolver.cpp
+++ b/src/Hydrodynamics/RiemannSolver.cpp
@@ -238,7 +238,7 @@ void ExactRiemannSolver<ndim>::ComputeStarRegion
     else if (2.0*fabs(pstar - pold)/(pstar + pold) < tolerance) break;
 
     // Check the star variables have not become NaNs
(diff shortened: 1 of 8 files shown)


src/Hydrodynamics/RiemannSolver.cpp
```diff
@@ -238,7 +238,7 @@ void ExactRiemannSolver<ndim>::ComputeStarRegion
     else if (2.0*fabs(pstar - pold)/(pstar + pold) < tolerance) break;
 
     // Check the star variables have not become NaNs
-    if (pstar != pstar || ustar != ustar) {
+    if (pstar != pstar) {
```
giovanni-rosotti: What is this doing in this branch? What does it have to do with DeleteDeadParticles?

rbooth200:
It doesn't really.

But we were getting some travis failures from ustar not being initialized. This was appearing in shocktube tests where the exact riemann solver is used for the python solutions.

I could have put it somewhere else, but it caused a travis failure on this branch in Jan, so I fixed it here thinking that it would get merged quickly...

PR #129, opened 2017-01-24
Unused/Uninitialised variables

dhubber:
Cleans up many compiler warnings by : 
- Removing unused variables
- Setting uninitialised variables to zero/some initial value (may be simply a warning but could potentially be actual bugs)
- Cast some unsigned int variables to ints to prevent int/unsigned comparison warnings (used C-style casts; these are mainly from STL size/length functions that return unsigned ints)

This also includes the uninitialised variables in the Riemann solver that were causing issues before.  I've tested it on a sample of the test suite so now will see how it performs with Travis to check everything else.  

There are still two main files/modules that are producing mass warnings, namely the Multiple Source Radiation class and the TreeRay class.  Both of these will be addressed in separate commits in the future.

There was one final warning I was not sure what it was doing exactly so left it alone.  This is a warning regarding some 'reinterpret_cast<SNAPFLOAT&>' statements in SphSnapshot.cpp.  I'll leave this one alone to whoever wrote this code!

### DIFF
--- a/src/Common/Dust.cpp
+++ b/src/Common/Dust.cpp
@@ -532,7 +532,6 @@ void DustSphNgbFinder<ndim, ParticleType>::FindNeibAndDoForces
     int cc;                                      // Aux. cell counter
     int i;                                       // Particle id
     int j;                                       // Aux. particle counter
-    int jj;                                      // Aux. particle counter
     int k;                                       // Dimension counter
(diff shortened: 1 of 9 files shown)


dhubber:
Will replace soon with new pull request with up-to-date branch (from master) with cherry-picked commits.

PR #130, opened 2017-03-14
Added locally isothermal eos, refactored eos interface

rbooth200:
I've added a locally isothermal eos and refactered the interface of the eos classes to make life easier.  I meant to do this on master, but it ended up on paper tests.

Can you check the implementation and merge into paper tests if everything is ok...


dhubber:
I don't really have any experience with locally isothermal discs so not sure how to check that really, but I'll check that the other EOSs are still fine and working as expected before merging.

rbooth200:
Its something that everybody would probably end up modifying for their
particular application, but I hope that it's a useful skeleton at least...

dhubber:
Okay, I checked some of the other EOSs and they all seem fine.  This branch is failing more Travis tests than final_paper_tests but I don't think it's because of any new bugs but some memory leak issue with the Ewald forces table (which we need to solve quickly).  So unless there are any objections, I will merge this into final_paper_tests and then look at beginning to investigate that issue from there.

rbooth200:
Fine by me.

rbooth200:
By the way I just noticed that I had a (small) commit that I'd not pushed. I've now added it to final_paper_tests directly

PR #131, opened 2017-03-21
Periodic bug

rbooth200:
I've fixed a bug related to the neighbour search when periodic boundaries are used. In doing this, I have used the neighbour manager to reinstate the validation of the neighbour lists, which is working nicely where it is applicable.


rbooth200:
Closing this because it has some commits that shouldn't be here

PR #137, opened 2017-03-28
Nbody fixes and paper figure scripts

dhubber:
Small pull request with some small fixes and modifications to the N-body parts, including : 
- Fix for identifying binary stars (was using centre of position instead of centre of mass)
- Parallelised intensive parts of N-body with OpenMP finally

Also added files and scripts to generate star-only and hybrid plummer spheres, run their simulations and generate figures for the paper.

### DIFF
--- a/src/Headers/Nbody.h
+++ b/src/Headers/Nbody.h
@@ -57,7 +57,17 @@ class Hydrodynamics;
 template <int ndim>
 class Nbody
 {
- public:
+protected:
+
+#if defined _OPENMP
+  static const int maxNbodyPerThread = 16;     ///< Max. no. of N-body particles per OpenMP thread
+                                               ///< if conditional OpenMP is employed.
+  const int maxNbodyOpenMp;                    ///< Max. total of N-body particles for OpenMP
+                                               ///< if conditional OpenMP is employed.
+#endif
+
+
+public:
 
   // Constructor and destructor functions
   //-----------------------------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


dhubber:
Also I added conditional OpenMP parallelisation to only use OpenMP when a reasonably high number of particles per thread are being simulated (to prevent OpenMP being excessively deployed when only a small number of stars/sinks are present).

rbooth200:
Looks ok to me. I have one question though: does it not make more sense to set the maximum number of OpenMP threads used rather than switch it on or off. Given the work is N^2 you could do nMaxThread = 1 + Npart^2 / 16^2 or similar...

dhubber:
> Looks ok to me. I have one question though: does it not make more sense to set the maximum number of OpenMP threads used rather than switch it on or off. Given the work is N^2 you could do nMaxThread = 1 + Npart^2 / 16^2 or similar..

Hmmm, perhaps.  I just went with the solution I was thinking of at the time but it's easy enough to change since it's only a few lines really.

giovanni-rosotti:
Hi, this looks mostly fine to me. Isn't more natural to call maxNbodyOpenMp  minNbodyOpenMp (it's the minimum number of particles needed to use openMP)?

David, are you going to do what Richard suggested, or do you think it's not worth the effort?

dhubber:
> David, are you going to do what Richard suggested, or do you think it's not worth the effort?

Oh well, it's an improvement but not essential (i.e. it's not a bug fix).  But it also kind of applies to other OpenMP parts of the code (e.g. when there are only a few active particles/cells) so I'd be tempted to say make an Issue (for the whole code) and look at it later.

PR #140, opened 2017-04-11
Fixed python call to Riemann solver with isotheraml eos

rbooth200:
This pull request fixes the analytical solution for shock tube problems that use an iosthermal eos.

The problem was the pressure being passed to the riemann solver was not consistent with the ones used in the simulation.


dhubber:
Great!  That worked thanks.

PR #142, opened 2017-05-03
Scott ionisation

dhubber:
Successfully merged Scott's ionisation algorithm with the latest version of the master branch.   Compiles successfully both in serial and OpenMP (not checked MPI but actually the algorithm won't work at all in MPI anyway so perhaps that option should be disabled, i.e. throw an exception).

Important things to do before merging with master : 
- Prepare and run tests of this algorithm (atm, only really have the Spitzer expansion test since it's a 3D only algorithm)
- Find acceptable error norms of a solution and prepare a nose2/Travis script to automate a test (low-res so it doesn't slow Travis down).  Btw, is there a way to make Travis skip tests in certain conditions?  This won't work in MPI so any MPI tests need to be skipped?
- One last clean-up of the code (plenty of unnecessary variables, operations and array allocations to optimise and better comments required).


rbooth200:
Did you mean to add IC.cpp back in?

dhubber:
> Did you mean to add IC.cpp back in?
 
No, it was in the commit for Scott's algorithm (which is almost a year old now) and I didn't spot it.  I'll get rid of it in the next commit.

dhubber:
ok, I've fixed the main small issues with Scott's routine and it seems to run fine now.  I've still to fully implement a solution in the python files that will allow an automatic nose/Travis run so this will be my aim in the next commit.  But at least a direct comparison can be made with this and Jim's implementation in the meantime.

PR #144, opened 2017-05-12
Mfv dust2

rbooth200:
This branch updates the dust implementation, introducing several improvements:
 - Both KDK and DKD leapfrog are now properly implemented with SPH
 - The dust now works with the meshless
 - The dust is now fully conservative.

To do this I have had to make a number of modifcations along the way, hence such a big pull request. Specifically:
 - The new timestep is now computed at the end of the old one, rather than the beginning of the new one. When implementing this I removed duplicate copies of the ComputeBlockTimesStep and ComputeGlobalTimeStep methods. This has meant introducing a new class into the meshless. The particles now store both their old and new timestep, because this is needed for correctly computing the kicks.
-  Allow the MPI ghost particles to transmit information back to the parent processors. 

I think these changes are now ready to go.


src/Hydrodynamics/EnergyRadws.cpp
```diff
@@ -207,22 +208,21 @@ EnergyRadws<ndim,ParticleType>::~EnergyRadws()
...
-  const int Npart,                     ///< [in] Number of particles
   const FLOAT t,                       ///< [in] Current simulation time
   const FLOAT timestep,                ///< [in] Base timestep value
-  Particle<ndim>* part_gen)            ///< [inout] Pointer to SPH particle array
+  Hydrodynamics<ndim>* hydro)
 {
   int i;                               // Particle counter
   FLOAT dt;                            // Timestep since start of step
-  ParticleType<ndim>* partdata = static_cast<ParticleType<ndim>* > (part_gen);
+  ParticleType<ndim>* partdata = hydro->template GetParticleArray<ParticleType>();
```
giovanni-rosotti: This is evil. What kind of template magic is going on?!

rbooth200: Nothing.  For some reason the standar requires template keyword when calling a template function with a template parameter that is itself a template parameter.

The motivation here is that GetParticleArray can now do some sanity checking on the requested type.

src/Mpi/MpiControl.cpp
```diff
@@ -589,35 +590,118 @@ void MpiControlType<ndim, ParticleType>::UpdateMpiGhostParents
...
+
+  MPI_Datatype MpiReturnType;
+  MPI_Datatype types[1] = {MPI_BYTE};
+  MPI_Aint offsets[1] = {0};
+  int blocklen[1] = {sizeof(ReturnDataType)};
 
-  MPI_Alltoallv(&buffer_iorig[0], &N_updates_per_proc[0], &displs_ghosts[0], MPI_INT,
-                &buffer_receive_iorig[0], &N_updates_from_proc[0], &displs_ghosts_receive[0],
-                MPI_INT, MPI_COMM_WORLD);
+  MPI_Type_create_struct(1,blocklen,offsets,types,&MpiReturnType);
```
giovanni-rosotti: I am not sure we want to create a MPI data type every time we call this function

rbooth200: Can you suggest a change?

rbooth200: OK done, about to push an update.

dhubber:
> Both KDK and DKD leapfrog are now properly implemented with SPH

Oh, was there a bug in the old implementation since it seemed to be working fine with me (or at least the KDK since I was using that one mainly)?  Or do you mean it's all consistent with your new changes to the integration schemes?


> To do this I have had to make a number of modifcations along the way, hence such a big pull request

Yes, it might be a bit tricky to review everything (although I can see many are effectively replace alls like the set/unset flag commands), especially since I don't really know much about the dust code.  I think seeing as it passes all the Travis tests, I'll just run a few tests manually just to satisfy myself and then I'll be fine with merging (although I'll wait for Giovanni's okay also).

rbooth200:
Before only the DKD dust was implemented correctly. If you ran KDK you'd probably still get something sensible because the code would have still reproduced the correct limits of very small and very long stopping times. However, in the middle it was formally incorrect.

Other than that the changes were _mostly_ straight swaps. The block time step code has actually been changed slightly: it now only computes the timestep for the next step, leaving the user to select it when ready...

dhubber:
> Before only the DKD dust was implemented correctly. If you ran KDK you'd probably still get something sensible because the code would have still reproduced the correct limits of very small and very long stopping times. However, in the middle it was formally incorrect.

Okay, so you're talking just about the dust algorithm.  No problem then!

I had a short look at the code and run some (non-dust) tests myself manually just for peace of mind.  If everyone else is happy, we can merge asap.

giovanni-rosotti:
Hi both,
the dust had some horrible bug with the sink accretion invalidating the tree when producing a snapshot at the end of the main loop (because of the call to DeleteDeadParticles). I've corrected this for SPH, and it seems to me that the meshless is already fine (please check!), because it always rebuilds the tree after sink accretion.

David, you should know that this makes the tree invalid when calling the supernova driver routine at the start of the following timestep. I don't know if that routine uses the tree, or if it ever will, but this is something you should be aware of.

rbooth200:
The meshless looks of RE the dust as far as I can tell.

src/Common/Dust.cpp
```diff
@@ -933,7 +933,7 @@ int DustInterpolant<ndim, ParticleType, StoppingTime, Kernel>::DoInterpolate
     // list error in order to generate a larger neighbour list (not properly implemented yet).
     if (h > hmax) return 0;
 
-  } while (h > h_lower_bound && h < h_upper_bound);
```
dhubber: This same inequality is also in the main grad-h (and maybe meshless?) h-finding also.  Could it cause a problem there also, or was there a reason it was bad news here?  I'll have a look at it in any case.

dhubber:
> David, you should know that this makes the tree invalid when calling the supernova driver routine at the start of the following timestep. I don't know if that routine uses the tree, or if it ever will, but this is something you should be aware of.

Yes, unfortunately it does call the tree when injecting supernova (since it needs to know surrounding neighbouring particles in case they need to be heated up also).  And then once the new particles are injected, it makes sense to immediately rebuild the tree with the new particles included.  So this definitely is a problem then since it was called from the ideal place before and now that doesn't work I guess.  I might have to figure out if it can be moved easily, perhaps before the sink particles are called?

rbooth200:
I'm not sure its such a problem for the SPH. The reason it arises with the dust is that dust is much more likely to enter patholigical particle arrangements, finding far too many or no neighbours. So I added some extra logic to help when no neighbours are found in an iteration, i.e. set h = hmax and iterate, but that violated the clause...

giovanni-rosotti:
I noticed that restarts aren't working with the dust (at least, after restart the timestep goes to 1e-10, so something crazy has happened!). It might not be the whole story, but I think the code we have now is clearly wrong. When we write a snapshot, we just record how many particles are gas and how many are dust. Then when we read the snapshot back in, we just say that the first Ngas particles are of gas type, and remaining of dust type. Now, this is correct if the initial conditions have been set up in this way, and if the order of the particles is maintained. But since there is accretion, this is not the case...
Can't we simply write ptype in the snapshot file? Is there any reason why we are not doing it already?

rbooth200:
I think the reason is just that I never thought about what happens with sinks. I would have no problem writing the ptype, but I think we already have a solution when using MPI, which, if I remember correctly, involves sorting the particles by type when writing them to disc. Would this be too difficult? 

I suspect that if we just write the ptype we are going to break everybody's seren readers...

rbooth200:
Ok, I so the read/write issue has been fixed in another branch. Is there anything else left that needs doing before this one can be merged?

dhubber:
> I suspect that if we just write the ptype we are going to break everybody's seren readers...

Actually it's designed to skip through any unrecognised arrays (as long as it's formatted correctly) in both gandalf and seren, so this should not be a problem to add an extra int array.  But if the fix you've recently made has already accounted for this, then no need to do it anymore I guess!

dhubber:
I just fixed some conflicts with master (and fixed a small other bug with VERIFY_ALL).  I guess this is ready to go then, unless there's something we've discussed about adding that I've forgotten about??

rbooth200:
I think this is ready to go, too. 

It might be easier if we merge the visc pull request first, to avoid any potential git issues...

dhubber:
ok, I see you've merged this one into visc so guess we can close this pull request and delete mfv_dust2 now?  I'll just wait until Travis gives the green light!

rbooth200:
No, merge the visc pull request and then the mfv dust one.

PR #146, opened 2017-05-18
Visc

rbooth200:
This branch, based off of the mfv_dust2 pull request implements two changes to the viscosity:

1) An addition of physical viscosity in the meshless, to add viscosity where desired
2) An implementation of the Cullen & Dehnen switch in the SPH, to remove viscosity where not desired.

When the mfv_dust2 branch is ready to go then this one should be too.

### DIFF
--- a/src/GradhSph/GradhSph.cpp
+++ b/src/GradhSph/GradhSph.cpp
@@ -140,15 +140,10 @@ void GradhSph<ndim, kernelclass>::DeallocateMemory(void)
 //=================================================================================================
 template <int ndim, template<int> class kernelclass>
 int GradhSph<ndim, kernelclass>::ComputeH
- (const int i,                         ///< [in] id of particle
-  const int Nneib,                     ///< [in] No. of potential neighbours
-  const FLOAT hmax,                    ///< [in] Max. h permitted by neib list
-  FLOAT *m,                            ///< [in] Array of neib. masses
-  FLOAT *mu,                           ///< [in] Array of m*u (not needed here)
-  FLOAT *drsqd,                        ///< [in] Array of neib. distances squared
-  FLOAT *gpot,                         ///< [in] Array of neib. grav. potentials
-  SphParticle<ndim> &part,             ///< [inout] Particle i data
-  Nbody<ndim> *nbody)                  ///< [in] Main N-body object
+ (SphParticle<ndim> &part,                                ///< [inout] Particle i data
+  FLOAT hmax,                                             ///< [in] Maximum smoothing length
+  const vector<DensityParticle> &ngbs,                    ///< [in] Neighbour properties
+  Nbody<ndim> *nbody)                                     ///< [in] Main N-body object
 {
   int j;                               // Neighbour id
   int k;                               // Dimension counter
@@ -170,6 +165,8 @@ int GradhSph<ndim, kernelclass>::ComputeH
     if (hmax < hmin_sink) return -1;
   }
 
+  int Nneib = ngbs.size();
+
   // Some basic sanity-checking in case of invalid input into routine
   assert(Nneib > 0);
   assert(hmax > (FLOAT) 0.0);
@@ -177,6 +174,7 @@ int GradhSph<ndim, kernelclass>::ComputeH
   assert(parti.m > (FLOAT) 0.0);
 
   FLOAT invh ;
+
 
   // Main smoothing length iteration loop
   //===============================================================================================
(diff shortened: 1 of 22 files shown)


rbooth200:
I changed the base to mfv_dust2, so you can see just the new changes in this branch...

dhubber:
Oh, this is very interesting and useful (particularly the viscosity for the meshless).  Is this being rushed through so we can also add it in the paper?  If so, we'll have to expand some of the tests a little, such as extra shocktube tests for the CD viscosity.

rbooth200:
Yes, I agree that this needs testing. It should be easy enough to add a nose test for the CD test tube. I'll do that soon... 

But I think giovanni plans to explicitly test them for the paper using the spreading ring test. I guess we could also add a nose test for the spreading ring, but Giovanni would need to add his 'analytical solution' for comparion.

src/GradhSph/GradhSph.cpp
```diff
@@ -207,7 +207,7 @@ int GradhSph<ndim, kernelclass>::ComputeH
 
     // Density must at least equal its self-contribution
     // (failure could indicate neighbour list problem)
-    assert(parti.rho >= parti.m*parti.hfactor*kern.w0_s2(0.0));
+    assert(parti.rho >= 0.99*parti.m*parti.hfactor*kern.w0_s2(0.0));
```
dhubber: I've had some bug-reports regarding this line when the computed density is exactly equal to the self-contribution.  It suggests some other problem (think it's to do with the sinks which I've solved in another branch) but also it could be the guess is simply too small so there's only the self-contribution, so I think making this slightly smaller than 1 is fine to stop this assert killing the code when there's not a problem.

src/GradhSph/GradhSph.cpp
```diff
@@ -140,15 +140,10 @@ void GradhSph<ndim, kernelclass>::DeallocateMemory(void)
...
- (const int i,                         ///< [in] id of particle
-  const int Nneib,                     ///< [in] No. of potential neighbours
-  const FLOAT hmax,                    ///< [in] Max. h permitted by neib list
-  FLOAT *m,                            ///< [in] Array of neib. masses
-  FLOAT *mu,                           ///< [in] Array of m*u (not needed here)
-  FLOAT *drsqd,                        ///< [in] Array of neib. distances squared
-  FLOAT *gpot,                         ///< [in] Array of neib. grav. potentials
-  SphParticle<ndim> &part,             ///< [inout] Particle i data
-  Nbody<ndim> *nbody)                  ///< [in] Main N-body object
+ (SphParticle<ndim> &part,                                ///< [inout] Particle i data
```
dhubber: Yes, probably a nicer and cleaner way to do this, and makes it easier if we need additional quantities for computing other things without needing to change the interface all the time.

rbooth200:
The problem with this line has been that floating point errors seems to
cause the assert to fail when it should pass with equality. That's why
we've been replacing it with the 0.99, after which the iteration over h and
neighbours takes care of it.

src/GradhSph/GradhSphTree.cpp
```diff
@@ -266,13 +247,6 @@ void GradhSphTree<ndim,ParticleType>::UpdateAllSphProperties
     //=============================================================================================
 
     // Free-up all memory
-    delete[] r;
```
dhubber: I'm guessing the memory allocated in the ngb/ngb2 vectors are automatically deallocated when their destructor is called? So no need to manually deallocate them here?

src/GradhSph/GradhSph.cpp
rbooth200: The problem with this line has been that floating point errors seems to
cause the assert to fail when it should pass with equality. That's why
we've been replacing it with the 0.99, after which the iteration over h and
neighbours takes care of it.

src/Headers/Particle.h
```diff
@@ -240,6 +240,31 @@ struct SphParticle : public Particle<ndim>
...
   }
+
+  class DensityParticle {
+  public:
+    DensityParticle() : m(0), u(0), gpot(0), ptype(0) {} ;
+    DensityParticle(const SphParticle<ndim>&p) {
+      for (int i=0; i<ndim; i++) {
+        r[i] = p.r[i];
+        v[i] = p.v[i];
+        a[i] = p.a[i];
```
dhubber: Is the acceleration needed here in this data structure?  Is it needed for example in the CD2010 viscosity?

src/GradhSph/GradhSphTree.cpp
rbooth200: Yes, that's correct. It's a big advantage of vectors

src/Headers/Particle.h
rbooth200: The velocity and acceleration are both needed for the CD2010 viscosity.

src/SM2013/SM2012Sph.cpp
```diff
@@ -247,14 +250,14 @@ int SM2012Sph<ndim, kernelclass >::ComputeH
     for (j=0; j<nbody->Nstar; j++) {
       invhsqd = pow(2.0 / (parti.h + nbody->stardata[j].h),2);
       for (k=0; k<ndim; k++) dr[k] = nbody->stardata[j].r[k] - parti.r[k];
-      ssqd = DotProduct(dr,dr,ndim)*invhsqd;
```
rbooth200: I removed the ssd variable, so i needed to comment it out. 

I'd rather not delete the SM2012 as it is so close to working. With a little effort I think we can get it working again without the need for code duplication. It's just not a priority at this second...

rbooth200: Surely the gravitational terms are needed with SM2012, just slightly different?

dhubber:
Okay, I've had a look through the branch and also tried it out on a couple of simulations.  However, I didn't really see much difference at all between the MM97 and CD2010 switches.  I only really tried a shocktube test (where perhaps I should not expect a big difference) and the shear flow test (where I would perhaps expect a bigger difference).  Or maybe these tests are not challenging enough to show any differences?  I'm guessing you're preparing some more sophisticated tests for both the CD2010 and the physical viscosity cases?  Otherwise there were no problems and the coding seemed fine so I'm happy with it.

rbooth200:
I agree that (hopefully) there should be little difference in the shock tube tests. Normally I'd use the gresho when exploring viscosity. I'll have a look later.

PR #147, opened 2017-05-18
Corrected writing of file format with dust when sinks are present for…

rbooth200:
… su and sf formats

The particles are now always sorted by type in the su/sf formats. I haven't changed the column format since it isn't supporring different particle types anyway.

### DIFF
--- a/src/MeshlessFV/MeshlessFVTree.cpp
+++ b/src/MeshlessFV/MeshlessFVTree.cpp
@@ -578,7 +578,7 @@ void MeshlessFVTree<ndim,ParticleType>::UpdateGodunovFluxes
             neibmanager.GetParticleNeib(activepart[j],hydromask,do_pair_once);
 
 #if defined(VERIFY_ALL)
-        neibmanager.VerifyNeighbourList(i, Nhydro, mfvdata, "all");
+        neibmanager.VerifyNeighbourList(i, mfv->Nhydro, mfvdata, "all");
 #endif
 
         // Compute all neighbour contributions to hydro fluxes
--- a/src/Tree/HydroTree.cpp
+++ b/src/Tree/HydroTree.cpp
@@ -1922,6 +1922,7 @@ void HydroTree<ndim,ParticleType>::CheckValidNeighbourList
     for (j=0; j<Ntot; j++) {
       for (k=0; k<ndim; k++) dr[k] = partdata[j].r[k] - partdata[i].r[k];
       drsqd = DotProduct(dr,dr,ndim);
+      if (partdata[j].flags.is_dead()) continue;
       if (drsqd <= kernrangesqd*partdata[i].h*partdata[i].h) trueneiblist[Ntrueneib++] = j;
     }
   }
@@ -1929,6 +1930,7 @@ void HydroTree<ndim,ParticleType>::CheckValidNeighbourList
     for (j=0; j<Ntot; j++) {
       for (k=0; k<ndim; k++) dr[k] = partdata[j].r[k] - partdata[i].r[k];
       drsqd = DotProduct(dr,dr,ndim);
+      if (partdata[j].flags.is_dead()) continue;
       if (drsqd < kernrangesqd*partdata[i].h*partdata[i].h ||
           drsqd < kernrangesqd*partdata[j].h*partdata[j].h) trueneiblist[Ntrueneib++] = j;
     }
--- a/src/Tree/HydroTree.cpp
+++ b/src/Tree/HydroTree.cpp
@@ -1018,7 +1018,7 @@ void HydroTree<ndim,ParticleType>::UpdateHydroExportList
     //=============================================================================================
 
 #ifdef VERIFY_ALL
-    PrintArray("Ncellexport : ",Nmpi,Ncellexport);
+    //PrintArray("Ncellexport : ",Nmpi,Ncellexport);
     PrintArray("Npartexport : ",Nmpi,Npartexport);
 #endif
 
--- a/src/Common/SimulationIO.hpp
+++ b/src/Common/SimulationIO.hpp
@@ -1761,8 +1761,8 @@ bool Simulation<ndim>::WriteSerenUnformSnapshotFile(string filename)
   types[2].ptype = cdm_type ;
   types[3].ptype = dust_type;
 
-  for (int i=0; n<hydro->Nhydro; i++) {
-    Particle<ndim>& part = hydro->GetParticlePointer(i);
+  for (int i=0; i<hydro->Nhydro; i++) {
+    const Particle<ndim>& part = hydro->GetParticlePointer(i);
     if (part.flags.is_dead()) continue;
     switch (part.ptype) {
       case icm_type:
@@ -1774,7 +1774,7 @@ bool Simulation<ndim>::WriteSerenUnformSnapshotFile(string filename)
       case dust_type:
         types[3].Ntot_type++; break;
       default:
-        ExceptionHandler::getIstance().raise("SerenUnformReader: Type not recognised");
+        ExceptionHandler::getIstance().raise("SerenUnformReader: Type not recognised : " + part.ptype);
     }
   }
 
--- a/src/Common/SimulationIO.hpp
+++ b/src/Common/SimulationIO.hpp
@@ -1699,7 +1699,7 @@ bool Simulation<ndim>::WriteSerenUnformSnapshotFile(string filename)
     typedata[ndata][4] = 0; ndata++;
 
     data_id[ndata] = "r";
-    typedata[ndata][0] = ndim; typedata[ndata][1] = 1;problem
+    typedata[ndata][0] = ndim; typedata[ndata][1] = 1;
     typedata[ndata][2] = Ntot_hydro; typedata[ndata][3] = 4;
     typedata[ndata][4] = 1; ndata++;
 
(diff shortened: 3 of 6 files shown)


dhubber:
So, does this now skip over any dead/accreted particles when writing the files?  What we'd like to do of course is remove that call to DeleteDeadParticles after sink accretion only for writing files (the one Giovanni cited) which would break other things.  I was thinking of a plan to get rid of this myself, but if this does it already, then great!!  You mention dust with sinks explicitly, but didn't know if this also included sink accretion in general.

rbooth200:
Actually, it doesn't skip the dead particles, but I can certainly add it in... I'll do that later

dhubber:
Okay, I think I can see how to skip the dead particles if you've not done it already (and if you have please push).  It'll require printing out the number of live hydro particles (not just Nhydro) and then skipping in these new 'Write' functions if the dead flag is active.  I can also add in writing and reading the ptype while I'm at it!

dhubber:
Actually, skip writing and reading ptype since the way you've added seems to do the trick and will simply overwrite any values anyway!  So guess there's no need for that anymore!

rbooth200:
Sounds good. If you have time please do it, i haven't got there yet...

dhubber:
Okay, I've implemented the main parts of ignoring dead particles.  It seems to run okay with sinks (during the Boss-Bodenheimer test with Grad-h SPH) apart from some problem that I've already fixed on another branch (and will make a pull request soon).  A few extra things to do : 
- double-check it runs fine without this call to DeleteDeadParticles in the MainLoop (so allowing us to hopefully merge mfv_dust2 also)
- should be tested with MPI also since I only tried OpenMP
- some multi-species simulation (e.g. gas and dust with sinks)

dhubber:
ok, MPI clearly not yet working with Travis.  Will look at that now.

src/MeshlessFV/MeshlessFVTree.cpp
```diff
@@ -578,7 +578,7 @@ void MeshlessFVTree<ndim,ParticleType>::UpdateGodunovFluxes
             neibmanager.GetParticleNeib(activepart[j],hydromask,do_pair_once);
 
 #if defined(VERIFY_ALL)
-        neibmanager.VerifyNeighbourList(i, Nhydro, mfvdata, "all");
+        neibmanager.VerifyNeighbourList(i, mfv->Nhydro, mfvdata, "all");
```
rbooth200: Does this need to be activelisy[j] also?

src/Tree/HydroTree.cpp
```diff
@@ -1922,13 +1922,15 @@ void HydroTree<ndim,ParticleType>::CheckValidNeighbourList
     for (j=0; j<Ntot; j++) {
       for (k=0; k<ndim; k++) dr[k] = partdata[j].r[k] - partdata[i].r[k];
       drsqd = DotProduct(dr,dr,ndim);
+      if (partdata[j].flags.is_dead()) continue;
```
rbooth200: Can you add these checks to the version of this function in the neighbour manager?

src/Tree/HydroTree.cpp
```diff
@@ -1922,13 +1922,15 @@ void HydroTree<ndim,ParticleType>::CheckValidNeighbourList
...
       if (drsqd <= kernrangesqd*partdata[i].h*partdata[i].h) trueneiblist[Ntrueneib++] = j;
     }
   }
   else if (neibtype == "all") {
     for (j=0; j<Ntot; j++) {
       for (k=0; k<ndim; k++) dr[k] = partdata[j].r[k] - partdata[i].r[k];
       drsqd = DotProduct(dr,dr,ndim);
+      if (partdata[j].flags.is_dead()) continue;
```
rbooth200: Same as above

rbooth200:
This is looking good as long as it passes the travis tests. However, I do think its necessary to make the column format explicity skip the dead particles like we are doing with the other format...

src/MeshlessFV/MeshlessFVTree.cpp
dhubber: Not sure what you mean.  This is in the active loop and i = activelist[j] already.  Or do you mean replace i with activelist[j]?

rbooth200: In which case I'm sure it's fine. I only mentioned it because you'd made similar changes elsewhere. Were they really needed?

dhubber:
> This is looking good as long as it passes the travis tests. However, I do think its necessary to make the column format explicity skip the dead particles like we are doing with the other format...

I seem to be getting some assertion failures in the HydroTree::FindMpiTransferParticles function (in the big debugging block near the end of the function).  Was thinking they were possible out-of-date/incorrect assertions but actually without them the code crashes with some memory/malloc error later.  So something else to fix!  :-/   Although so far, no problems with Travis.

Yes, the column format should also do this but just getting the difficult/important one out of the way!

src/MeshlessFV/MeshlessFVTree.cpp
dhubber: Yes, if only to allow it to compile with the VERIFY_ALL flag (guess it had not been compiled with this recently).

rbooth200: I guess not.

rbooth200:
> Yes, the column format should also do this but just getting the difficult/important one out of the way!

Ok

>I seem to be getting some assertion failures in the HydroTree::FindMpiTransferParticles function (in the big debugging block near the end of the function). Was thinking they were possible out-of-date/incorrect assertions but actually without them the code crashes with some memory/malloc error later. So something else to fix! :-/ Although so far, no problems with Travis.

As far as I can tell you haven't removed the delete dead particles from the main loop on yet, so I would expect it to work just fine here. Maybe you have some other local changes that are making the difference?

rbooth200:
Whats the status of this now?

dhubber:
Nothing more today sorry since it's a holiday in Germany and actually the
weather is nice for a change :-).   I was planning on looking at it later
when I'm back home.

dhubber:
ok, I've had a look this evening and still problems.  I just commented out the delete dead particles bit to check it with Travis but it's still failing asserts with the Boss-Bodenheimer test on my laptop.  In particular, it's failing on line 1519 in HydroTree.cpp (some verification block that's looking for edge cases).  Whoever wrote this bit perhaps can look at it and see if there's anything wrong, or it's catching a real problem?  If I switch off debug mode, it still eventually crashes in a different place.  But first things first.  I'll get back to my other tasks in the meantime.

rbooth200:
Was the bug introduced introduced in this branch, or is it in the master? If its in the master can you also check whether its present in the visc branch or mfv_dust2?

dhubber:
Sure, good shout, I'll check the other branches now since I didn't check them recently with MPI on my laptop.

dhubber:
ok, the master branch has the same assertion failure on the same line!  Will see if I get the same error in Linux now just in case it's a Mac thing!

rbooth200:
Ok. So I think this branch should be ok to merge. Can you remove the commented out code entirely? Perhaps Giovanni, should look over it first.

dhubber:
ok, just done that although this is only for SphSimulation (will need to do the Meshless also assuming it has a similar delete particle line).  I assume you mean Giovanni should look before we merge?

rbooth200:
> I assume you mean Giovanni should look before we merge?

Yep. 


I think we'll need to remove these lines again in the mfv_dust2 branch because I moved them around. But I can worry about that anyway. 

When you have time can you look at the visc branch?

dhubber:
> When you have time can you look at the visc branch?

Sure, you mean as in look at it with a view to merging rather than testing for this bug/issue in that branch also??

rbooth200:
The first...

giovanni-rosotti:
Ok trying to catch up with this long conversation... do I understand correctly that you are asking me to check at line 1519 of HydroTree.cpp?
The check seems correct to me. If it fails, it probably means we have a problem either with the tree, or with the domain boxes. That line is checking that if we have a particle outside of the local domain, we are exporting it to some other domain.
I am happy to debug this (although maybe it should happen in a separate branch). What are you running exactly?

dhubber:
> I am happy to debug this (although maybe it should happen in a separate branch). What are you running exactly?

Well, that's up to you but this branch is still pretty small so can be done here if you want.  I ran the Boss-Bodenheimer test (the one in tests/astro_tests should do fine) in MPI (no OpenMP) on just 2 cores on my laptop (and with COMPILER_MODE = DEBUG and DEBUG_LEVEL = 1).

rbooth200:
Since this is a bug in the master I think that this branch should be judged on its own merits, and David's issue tackled in a seperate branch

rbooth200:
I agree that (hopefully) there should be little difference in the shock tube tests. Normally I'd use the gresho when exploring viscosity. I'll have a look later.

giovanni-rosotti:
Do I understand correctly that this is ready to go? If yes, I am happy for it to be merged - the MPI issue deserves a separate branch and issue as said.

rbooth200:
I think it is

giovanni-rosotti:
Ok, I was about to merge but noticed that the meshless main loop still contains a call to DeleteDeadParticles. Is there a reason for that? If not, I'll get rid of that.
Btw, now the SPH main loop (and I guess the meshless as well very soon) does not call DeleteDeadParticles any more; it happens only at tree rebuild. I guess this is the best; in this way we don't break the tree by accident, and the main loop does not see such a low level operation.

rbooth200:
That's best. I think the motivation for not fixing the meshless was that we
will need to fix it again in the dust branch. But either way it's fine...

giovanni-rosotti:
In this case I'll accept this as it is.

PR #148, opened 2017-06-01
Sink nbody update

dhubber:
Several small updates for sinks :
- Added additional criteria for sink formation (long overdue!).  
- Fixed a bug related to sinks with conservative SPH gravity (i.e. Price & Monaghan 2007, Hubber et al. 2013) which can lead to particle ejections if there are only a couple of particles in a sink with a high relative mass..
- Changed h/rho iteration to use functions in grad-h iteration.  Useful for previous experiments with sinks with SPH changes (didn't work but useful to keep this way for clarity and might resurrect in the future).


dhubber:
Before, the entire zeta term (gas + stars) was zeroed but now only the star term is zeroed if switched off by the parameter.

rbooth200:
Its inside paper revisions

PR #151, opened 2017-06-07
Added back shear waves with isothermal Riemann solver

rbooth200:
I realized that the HLL Riemann solver does not correctly capture shear waves, so I've now implemented an HLLC like solver that has a constant density as well as pressure and velocity in the mid-state, but allows the transverse velocities to vary. This option is only switched on when the simulation is 'isothermal'. It should work for both the isothermal and locally isothermal EoS. 

While the solver is idealistically better, in some simple tests I can't see any difference...

### DIFF
--- a/src/Headers/RiemannSolver.h
+++ b/src/Headers/RiemannSolver.h
@@ -183,9 +183,8 @@ class HllcRiemannSolver
     else if (Smin >= 0)
       Hydro_Flux(Sl, n, flux) ;
     else {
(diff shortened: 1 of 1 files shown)


dhubber:
Why only isothermal?  Surely this is just as relevant for adiabatic and other equations of state?

rbooth200:
The shear waves were already present with the adiabatic solver, just missing from the isothermal one.

PR #152, opened 2017-06-12
Fix bug that diagnostics, timings and restart snapshots are no longer produced

giovanni-rosotti:
After Richard's modifications, n is now 0 rather than nresync when the routine "Output" is called


rbooth200:
Looks fine to me, merging

PR #155, opened 2017-06-27
Prevent to run the code with ntreebuildstep different from 1 if using MPI

giovanni-rosotti:


rbooth200:
I opened an issue (#156) regarding this in case it does not get fixed today.  Accpetiong the hot fix in  the mean time.

giovanni-rosotti:
You haven't actually accepted the hotfix :)
It fails on travis because the freefall tests have nbuildtree != 1. I'll set it to something else...

rbooth200:
I was waiting to see if it passed... I'll do that now

PR #157, opened 2017-06-29
Improve openmp2

rbooth200:
Let's merge all of these optimizations from the investiagtions during the paper. I'm going to rebase this (if possible) and roll back the hot-fix for MPI...


rbooth200:
Ok, I've rebased and removed the hotfix. Now we only need @giovanni-rosotti's  fix for FractionBoxOverlap

rbooth200:
I've added these fixes, it should now be ready to go...

src/Common/Dust.cpp
```diff
@@ -411,8 +439,14 @@ void DustSphNgbFinder<ndim, ParticleType>::FindNeibAndDoInterp
...
 #endif
 
-  debug2("[DustSphNgbFinder::FindNeibAndDoInterp]");
-  CodeTiming::BlockTimer timer = timing->StartNewTimer("DUST_GAS_INTERPOLATE_FORCES");
+#ifdef _OPENMP
+  int Nthreads  = omp_get_max_threads() ;
+#else
+  int Nthreads  = 1 ;
+#endif
+  for (int t = neibbuf.size(); t < Nthreads; ++t)
```
giovanni-rosotti, line 450: Is this a trick to do the initialization only the first time that this function is called?

src/Common/Dust.cpp
```diff
@@ -660,128 +636,129 @@ void DustSphNgbFinder<ndim, ParticleType>::FindNeibAndDoForces
...
-      }
-    }
-
 #pragma omp critical
-    {
-      for (i=0; i<hydro->Ntot; i++) {
-        dudt[i] += dudt_local[i] ;
+      {
+        for (i=0; i<hydro->Ntot; i++) {
+          dudt[i] += dudt_local[i] ;
```
giovanni-rosotti: This doesn't need to be critical, if we do like we have done in the other loops.
If we do it, then the two barriers afterwards are not needed anymore

src/Headers/GhostNeighbours.hpp
```diff
@@ -324,7 +325,7 @@ class GhostNeighbourFinder
 	{
 		if (_any_special){
 		  for (int k=0; k < ndim; k++){
-		    FLOAT dr_k = p.r[k] - cell.rcell[k] ;
+		    FLOAT dr_k = p.r[k] - 0.5*(cell.bb.min[k] + cell.bb.max[k]);  //cell.rcell[k] ;
```
giovanni-rosotti: Aargh! Isn't worth it making this a function?

src/Headers/GhostNeighbours.hpp
```diff
@@ -527,14 +531,16 @@ struct DomainCellProxy
...
+	//using TreeCellBase<ndim>::rcell ;
+  FLOAT rcell[ndim];
 
 	DomainCellProxy(const DomainBox<ndim>& box)
 	{
 		for (int k=0; k<ndim;k++)
 		{
 			rcell[k] = 0.5 * (box.max[k] + box.min[k]) ;
 			FLOAT dr = 0.6 * (box.max[k] - box.min[k]) ;
+      // DAVID : Why is this 0.6 and not 0.5 (or 0.5000001)??
```
giovanni-rosotti: Somebody ought to reply to this comment

src/Headers/Hydrodynamics.h
```diff
@@ -50,7 +51,7 @@ class EOS;
 
 class SimulationBase;
 
-static const FLOAT ghost_range = 2.5;
+static const FLOAT ghost_range = 3.0; //2.5;
```
giovanni-rosotti: Why was this increased? We already have far too many ghosts...

dhubber:
Well, I guess this is basically close to ready.  If it passes Travis, then I guess it's time to merge into master.  But who's going to do the honours since we've all contributed heavily to this pull request?  :-)

rbooth200:
Giovanni plans to go through this carefully first I think...

src/Nbody/Sinks.cpp
```diff
@@ -463,6 +461,29 @@ void Sinks<ndim>::AccreteMassToSinks
     //---------------------------------------------------------------------------------------------
 
 
+    // Determine the closest sinks to all particles that are inside sinks
+#pragma omp for schedule(dynamic,1)
```
giovanni-rosotti: This schedule is probably going to have a huge overhead. More importantly, I think we can remove all together a loop over all particles constructing a list of particles inside sinks (it will be small!)

dhubber:
Okay, looking at it, Giovanni made the fewest commits so perhaps it's better he does it to give most of the changes better scrutiny!

giovanni-rosotti:
Yes, I am looking at it. I'm almost finished actually...

giovanni-rosotti (commented):
I have only very few things...

src/Headers/GhostNeighbours.hpp
dhubber: Actually it is a function (in TreeCell but can't remember the name).  Maybe I changed this one before making the function and then forgot to go back and correct it, so it just slipped through the net.

src/Common/Dust.cpp
rbooth200: Yes, this is to avoid the memory allocations. It seemed easy enough to do it here rather than having a seperate allocation / deallocation function.

src/Headers/Hydrodynamics.h
dhubber: Travis was weirdly failing a test with the error norms, expecting 0.004 and getting 0.0040002 (or something like that).  I just hypothesised maybe it was a ghost issue and actually it 'solved' it (in the sense it was passing Travis).  But otherwise I don't know what was going on.  We can try again reducing ghost_range and see if we pass all the tests now (maybe there was a problem elsewhere that's now been fixed?)

src/Common/Dust.cpp
rbooth200: I'll do this.

src/Headers/GhostNeighbours.hpp
rbooth200: Ok, I can do that.

src/Nbody/Sinks.cpp
dhubber: Hmmm, maybe remove, or maybe have conditional parallelisation only if Nsinks is greater than some threshold?  Either way, I don't mind so much.

src/Headers/GhostNeighbours.hpp
rbooth200: This object is no longer used, it was from the old brute force implementation. I'll remove it.

src/Headers/Hydrodynamics.h
rbooth200: Yeah, I'd prefer reducing it to 2, probably...

src/Nbody/Sinks.cpp
giovanni-rosotti: No no, this loop is over Nhydro, this is the problem. Loops of order N_particles should be avoided!!!!!

giovanni-rosotti:
Btw, why isn't Travis running tests anymore?

rbooth200:
I think there was a problem with travis earlier. Its meant to be fixed now so lets see what happens when I push the changes.

src/Common/Dust.cpp
rbooth200: I've done this (needs checking though).  I think a barrier is still needed after the MPI communication.

rbooth200:
Ok, I've made the changes I'm responsible for. 

Travis seems to be working again, too.

src/Headers/Hydrodynamics.h
rbooth200: This issue with a small change in the error for the sedov test has appeared a few times and then dissappeared. It could be a ghost range issue, but I'd be surprised. I've seen it occur occaisonally, it was the origin for the failure in this [commit](https://github.com/gandalfcode/gandalf/pull/157/commits/069a6ea907bec4ca3206ab622da787ebcd80ef87).

src/Nbody/Sinks.cpp
dhubber: Yes, but it contains a loop over Nsinks which makes the order of the whole loop Nhydro*Nsinks.  However, looking at it further now,  it's also checking for the particles which are inside sinks, which will only likely be a small fraction.  Then I guess most of the time this loop just cycles over to the next particle so is probably best to change it.  So I guess what you're suggesting is to make a list of particles which have sinkid != -1 and then loop/iterate over that list only?  That sounds like a reasonable solution anyway.

src/Headers/Hydrodynamics.h
dhubber: Yes, I'd prefer it smaller for obvious reasons.  But I also don't like the Travis tests being so sensitive to failing like these ones.

src/Nbody/Sinks.cpp
dhubber: And btw of course, I can do this tomorrow!

src/Headers/Hydrodynamics.h
rbooth200: My point is I don't think this change has fixed the bug...

dhubber: Yes, I know.

dhubber:
ok, I've made the change to Sinks.cpp now.  Basically I've modified the first two loops and put them in a master block with only the larger section (that actually does most of the work) in the parallel dynamic block.  There are perhaps ways to improve this even more (e.g. moving the master block region out of the omp region entirely) but this addresses the immediate issue hopefully.

rbooth200:
Why don't we parallelise over the number of sinks using OpenMP, in the same way that we do over cells for the normal tree walks?

dhubber:
Are you talking about the accrete loop?  Because it does loop over sinks.
Or are you talking about other places like when computing gravity for sinks?

rbooth200:
The loop over sinks in the master region.

dhubber:
Because there was this line 'plist[Ninsink++] = i;' which I wasn't sure I could do with atomics or without criticals.  But I'll probably re-write it another way anyway in the future so it's not too important.

dhubber:
Is there any reason we've not merged this yet?  I know there was this strange issue with Travis marginally failing a test randomly, but do we have any reason to suspect the cause of this originated in this branch?  If not, then we should merge asap.  But Giovanni should still be the final judge of this of course!  :-)

giovanni-rosotti:
Sorry, I had taken some vacation from GANDALF! :)
I think there are two issues left:
a) the right value to use for ghost_range;
b) the parallelization for the sinks, that I need to catch up with.
I am happy to delay the latter to a separate issue, but I feel that we can decide the former immediately, it's not such a big decision I would say... is there any reason why we should NOT go back to 2.5?

rbooth200:
I'm happy with 2.5

rbooth200:
This has failed because there is a ~ 1 per cent random fluctuation in the error norm when running with MPI. The SPH tolerence in the error norm is very close to the actual error in the serial case, so it's causing random fails.

Thoughts?

giovanni-rosotti:
Do your recent changes to MPI change anything with respect to the small difference in the error norm?

dhubber:
> Do your recent changes to MPI change anything with respect to the small difference in the error norm?

If you mean my changes, they were for a separate issue but they push the error norm just above the threshold as before so it's probably that ghost range issue again.  Was this ghost range issue fixed/discussed on another branch/pull request by any chance?  If we don't know what the cause is, or how to fix it, the simplest 'fix' is to increase the maximum error norm value for Travis to pass these tests.  I don't like that solution particularly but not sure what else to suggest otherwise.

giovanni-rosotti:
No I meant Richard's recent changes (didn't know you were doing any modification to MPI, David!)

dhubber:
> No I meant Richard's recent changes (didn't know you were doing any modification to MPI, David!)

No, I wasn't doing any direct MPI modifications but the timesteps bug I fixed was a potential MPI bug (in the way the global min/max timesteps were reduced over MPI) so wasn't sure if you were talking about that.

rbooth200:
What do you mean by recent? Changing the ghost range doesn't solve the issue. 
Otherwise my changes in the MPI/Sinks branch only affect sink particles.

dhubber:
> Changing the ghost range doesn't solve the issue.

It doesn't solve the issue but does seem to affect the final error and push it just below the threshold and so passes the Travis tests.  So either the ghost range is actually not big enough or there's something else at play (probably the latter).  Still scratching my head for ideas to solve it.  On the other hand, if it's not an issue that's just with this branch (we had similar issues with other branches I remember), then shouldn't we find a way to finally merge this branch and then deal with this problem on a separate branch?

rbooth200:
Actually, I don't think that is true, the result is quite random after all. Sometimes it passes, sometimes it fails with both values of the ghost range

dhubber:
I've only ever seen it fail with the smaller value (2.5) though.  I agree it's not as simple as that (as I wrote above) and maybe it's a race condition, either in particle ordering of computing forces or an MPI summation?  Maybe with MPI it leads sometimes to another h compute leading to slightly different h's and therefore slightly different errors?  If it's simply this, the acceptable error tolerance is way to close to the computed tolerance and should be increased. But if it were the case, it would also be nice to try and guarantee reproduceable results (probably impossible with our MPI scheme though).

rbooth200:
I definitely remember seeing it fail with both. I don't have any evidence any more though due to a rebase...

rbooth200:
The MPI is always going to be a bit non-deterministic due to the load balancing, so small fluctuations have to be expected. I don't know how small though, and ~1% seems on the high side to me so I suspect that there might be a bug somewhere.

dhubber:
See issue https://github.com/gandalfcode/gandalf/issues/181

PR #162, opened 2017-08-29
cd2010 viscosity fix

RoguePotato:
A small fix which removes the zeroing of dalphadt before it is used by the Cullen & Dehnen (2010) viscosity method.


rbooth200:
See Issue #162. I'll accept this once the travis tests pass.

PR #163, opened 2017-09-06
Acknowledge richard

giovanni-rosotti:
Any other place I forgot?


giovanni-rosotti:
David, could you please accept this immediately (double checking that I remembered to acknowledge Richard in all the relevant places)? This is really something that takes only 30 seconds...

PR #166, opened 2017-10-03
MPI/Sinks bug fix

rbooth200:
The guys in preston noticed that MPI wasn't working properly, we tracked it down to only problems in which sinks were included. This pull request includes a series of fixes to these issues:

1) The sinks data structure was being re-allocated, overwriting existing data
2) Copy-back of the update over MPI was broken
3) Some debug code was missing an edge case (dead particles)
4) Stocking of the tree after the sink calculation failed with the meshless: A rebuild was happening, invalidating the MPI ghosts.
5) Smoothed accretion is not properly handled with the meshless and MPI (fix to come).


rbooth200:
I've now added the final fix, this branch should be ready to go.

dhubber:
ok, just to confirm, you said in your last comment this is ready to go but then added 4 more commits.  So is it ready now or is there more code to come?  Also, is there actually a Travis-test that tests the sinks with MPI (e.g. Boss-Bodenheimer)?  I think so right?  If there is and the coding is done, then I'm happy to merge it now.

rbooth200:
Yes, it is ready to go

I actually don't know why MPI with the boss bodenheimer wasn't failing, but it was most obvious with the disc test or binary accretion test (where the sinks are present at the beginning).

PR #169, opened 2017-10-06
Preventing overwrite of iorig on restart

rbooth200:
We were re-setting iorig on restarts, which I don't think is the correct behaviour. This now skips it.


dhubber:
Is there a particular reason the Sph and Meshless parts look different (since I'd expect them to be identical for this bit), or is that just how it evolved with the code diverging slightly?

rbooth200:
No idea to be honest. Both of these functions are awful, but they work so I'm not about to mess around with them.

PR #172, opened 2017-10-31
Mfv radws

rbooth200:
This PR builds on the mfv_eos branch to implement cooling into the meshless. I've done this in the framework of the energy integration / radws schemes used in the sph.

This has been done in the following way:
 - Implementing an implicit cooling function for stability
 - Changing how the kinetic energy change from the gravitational acceleration is included in the energy update. This is now done using delta_KE, rather than dE/dt. They are equivalent to truncation error, but the new scheme is more robust.

This PR is ready to go...


dhubber:
I'll wait until the other eos branch is merged with master before looking through this since I suspect most of the file changes are common to both.

rbooth200:
See #171

rbooth200:
This branch has failed, but it's actually just a different manifestation of the same intermittant bug that keeps appearing on improve_openmp2 branch. I'm going to re-run the test that failed to show this.

rbooth200:
Ok, I've sorted out the merge.

rbooth200:
Ok, the merge should now be fixed

PR #173, opened 2017-10-31
Meshless gravity fix

rbooth200:
This pull request fixes some errors relating to the meshless and self-gravity with the gadget mac. Before these changes atree was being zero'd in the initializaiton resulting in a brute-force gravity calculation. 

There was also a mistake for the imported cells, which has been fixed: we were trying to recompute the acceleration tolerance, amin, for imported cells, which does not work because atree is zerod.

### DIFF
--- a/src/Headers/CommunicationHandler.h
+++ b/src/Headers/CommunicationHandler.h
@@ -406,7 +406,6 @@ public:
     c.m = 0;
     c.hmax = 0;
     c.maxsound = 0;
-    c.amin = big_number;
     for (int k=0; k<ndim; k++) {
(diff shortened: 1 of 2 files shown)


dhubber:
Seems fine.  I guess this is all independent of the other Mfv/radws/eos pull requests so I can merge now?

rbooth200:
yes

giovanni-rosotti:
I am happy for this to be merged. I assume that the bug with brute force was only affecting the first step anyway?

rbooth200:
Yeah, I think so. I've got some simulations with the gadget mac running fine at the moment. They are certainly faster than brute force.

PR #177, opened 2017-11-22
Improve ics2

dhubber:
Following improvements/fixes already added : 
- Changed Soundwave test to allow 2d or 3d tests
- Changed discontinuity test to generate ICs similar to Hopkins (2013) square pressure-discontinuity test
- Removed any remaining GetValue calls (now using GetDensity for regularisation)

A few possible changes to come before this should be merged
- Clean-up parameters files to remove old/un-unused parameters (e.g. riemann_order, sph)
- Improve/clean-up some comments (plenty of copy-and-pasting that was not corrected)
- Add GetDensity and SetProperties function to some other IC classes to allow regularisation to be used (Giovanni, Richard : could the DiscIc class be changed to use the regularisation?  It starts off with random sampling which could instead be done by the other functions in the IC class and then regularisation could 'clean-up' the noise)

Any other suggested IC class improvements that could go with this pull request (but nothing too big obviously)?

### DIFF
--- a/src/Headers/Ic.h
+++ b/src/Headers/Ic.h
@@ -584,7 +584,7 @@ public:
   virtual ~GreshoVortexIc() {};
 
   virtual void Generate(void);
-  virtual FLOAT GetValue(const std::string, const FLOAT *);
+  virtual FLOAT GetDensity(const FLOAT *, const int) const;
 
 };
 
@@ -788,7 +788,6 @@ public:
   virtual ~PolytropeIc();
 
   virtual void Generate(void);
-  virtual FLOAT GetValue(const std::string, const FLOAT *);
   static void ComputeIsothermalLaneEmdenSolution(const int, const FLOAT, FLOAT *, FLOAT *, FLOAT *, FLOAT *);
   static void ComputeLaneEmdenSolution(const int, const FLOAT, const FLOAT, FLOAT *, FLOAT *, FLOAT *, FLOAT *);
 
--- a/src/Ic/BossBodenheimerIc.cpp
+++ b/src/Ic/BossBodenheimerIc.cpp
@@ -162,8 +162,8 @@ void BossBodenheimerIc<ndim>::Generate(void)
 
 
 //=================================================================================================
-//  BossBodenheimer::GetValue
-/// Returns the value of the requested quantity at the given position.
+//  BossBodenheimer::GetDensity
+/// Returns the density at the given position.
 //=================================================================================================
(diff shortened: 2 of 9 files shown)


rbooth200:
>Add GetDensity and SetProperties function to some other IC classes to allow regularisation to be used (Giovanni, Richard : could the DiscIc class be changed to use the regularisation? It starts off with random sampling which could instead be done by the other functions in the IC class and then regularisation could 'clean-up' the noise)

Regularization could be added, but I would not make the disc IC use other functions in the IC class. The disc IC class needs to draw particles from special distributions, I see no point in changing something that works and is a special case. There is, however, an analytic form for the density so you could use that for the regularization, if you wish.

dhubber:
> Regularization could be added, but I would not make the disc IC use other functions in the IC class. The disc IC class needs to draw particles from special distributions, I see no point in changing something that works and is a special case. There is, however, an analytic form for the density so you could use that for the regularization, if you wish.

Sure, if it's set-up to work then no need to change the random sampling.  But surely it would still work with the Monte-Carlo rejection sampling??  Technically it's slower because it has to repeat but it should give the same quality of ICs surely?  But yes, I'm mainly asking purely to enable regularisation to decrease the start-up noise.

Btw, there's some strange issue that's causing all the soundwave test cases to fail which I don't understand at all (it works fine on my laptop and the test is extremely simple) so I'll open a new branch and add my changes one by one to figure out why it's failing!

rbooth200:
This might be useful:
https://git-scm.com/docs/git-bisect

rbooth200:
>Sure, if it's set-up to work then no need to change the random sampling. But surely it would still work with the Monte-Carlo rejection sampling?? Technically it's slower because it has to repeat but it should give the same quality of ICs surely?

The current code is doing rejection sampling.

dhubber:
> This might be useful:
> https://git-scm.com/docs/git-bisect

Yes, I knew about this for the command line (and should use it more often), but can it be used in some way on git-hub with Travis that doesn't involve pushing from my laptop all the time?  It actually works perfectly fine on my laptop; it's only breaking when it goes through the test suite on Travis :-/

rbooth200:
Are you valgrind clean?

dhubber:
Hmmm, I don't know.  I don't often use Valgrind (since it's so bloody slow) but might be a good call to use it in this case.

dhubber:
Closing and moving effort to different branch.

PR #180, opened 2018-02-16
addressing nbdoy level problem

shashwata95:
To change the nbody level was not always at a higher level than hydro particles.

### DIFF
--- a/src/Common/Simulation.cpp
+++ b/src/Common/Simulation.cpp
@@ -1755,7 +1755,6 @@ void Simulation<ndim>::ComputeGlobalTimestep() {
 
 
 
-
 //=================================================================================================
 //  Simulation::ComputeBlockTimesteps
 /// Compute timesteps for all particles using hierarchical block timesteps.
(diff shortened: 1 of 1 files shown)


src/Common/Simulation.cpp
```diff
@@ -2171,11 +2175,11 @@ void Simulation<ndim>::ComputeBlockTimesteps(void)
   }
   if (nbody != NULL) {
     for (i=0; i<nbody->Nnbody; i++) {
-      assert(nbody->nbodydata[i]->level <= level_max);
+      assert(nbody->nbodydata[i]->level <= level_max); 
       assert(nbody->nbodydata[i]->nlast <= n);
-      assert(nbody->nbodydata[i]->nstep == pow(2,level_step - nbody->nbodydata[i]->level));
+      assert(nbody->nbodydata[i]->nstep >= pow(2,level_step - nbody->nbodydata[i]->level)); //comment 5: what failed then, was solved by adding inequality sign instead of equality
```
dhubber, line 2180: The original assertion (nstep == pow(2, ...) ) was correct.  If it's failing, then there is a problem somewhere.

dhubber:
I think most of the proposed changes here are correct (with one exception which I commented about).  If you can just clean up the commit, in particular remove the comments since we don't want them in the main branch, then I'm happy to accept the pull request.

